PR #4, opened 2024-03-22
Remove console.log

samselikoff:
Believe this was leftover, I'm seeing it in my app

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -30,7 +30,6 @@ async function updateSession(request: NextRequest) {
   const newRequestHeaders = new Headers(request.headers);
 
   if (hasValidSession) {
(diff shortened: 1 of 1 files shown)


PaulAsjes:
Thanks @samselikoff!

@benoitgrelard what if we put this and the other console logs behind a debug config parameter? For instance:

```js
// doesn't show logs (default)
const { user } = await getUser();

// shows logs
const { user } = await getUser({ debug: true });
```

 I think they could still be useful if you're debugging session management but a little noisy if you aren't.

benoitgrelard:
Yep, I'll take care of that. @samselikoff I'll close this one and will deal with this separately. Thanks!

PR #5, opened 2024-03-22
Add impersonation support

benoitgrelard:


src/impersonation.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { signOut } from './auth.js';
+import { workos } from './workos.js';
+import { Button } from './button.js';
+import { MinMaxButton } from './min-max-button.js';
+
+interface ImpersonationProps extends React.ComponentPropsWithoutRef<'div'> {
+  side?: 'top' | 'bottom';
+}
+
+export async function Impersonation({ side = 'bottom', ...props }: ImpersonationProps) {
```
andy-hook, line 12: What happens when imported in a non app router scenario?

benoitgrelard: I haven't had the time to explore this as this was all very last minute, hence why we are in unstable `0.x.x` land for now. We'll have a ton more stuff to explore potentially.

andy-hook (approved):

PR #9, opened 2024-03-26
Add additional configuration to control the return pathname

### LINKED ISSUE #11: Implement Automatic Configuration for Redirecting Users After Login
Is it feasible to implement automatic configuration for a parameter similar to 'afterSignInUrl,' ensuring that when a user attempts to access a protected page and then successfully logs in, they are redirected to the original page they were trying to reach?

For example:
1. A user tries to access a restricted page on a website, such as "example.com/secure-page".
2. They are redirected to the login page to authenticate (Authkit hosted page).
3. After successfully logging in, they are automatically redirected back to "example.com/secure-page" instead of being directed to a default page ("/")

benoitgrelard:
This PR handles a couple things:

- Changes the way the callback route is configured, to enable passing some options (notably an option to configure the default return pathname)
- Handles returning the user to the pathname they attempted before they were thrown to sign-in when `ensureSignedIn` is `true` in `getUser()`

Fixes #11

### DIFF
--- a/src/authkit-callback-route.ts
+++ b/src/authkit-callback-route.ts
@@ -4,10 +4,7 @@ import { workos } from './workos.js';
 import { WORKOS_CLIENT_ID } from './env-variables.js';
 import { encryptSession } from './session.js';
 import { cookieName, cookieOptions } from './cookie.js';
-
-interface HandleAuthOptions {
-  returnPathname?: string;
-}
+import { HandleAuthOptions } from './interfaces.js';
 
 export function handleAuth(options: HandleAuthOptions = {}) {
   const { returnPathname: returnPathnameOption = '/' } = options;
@@ -32,7 +29,7 @@ export function handleAuth(options: HandleAuthOptions = {}) {
         url.searchParams.delete('state');
 
         // Redirect to the requested path and store the session
-        url.pathname = returnPathname ? returnPathname : returnPathnameOption;
+        url.pathname = returnPathname ?? returnPathnameOption;
 
         const response = NextResponse.redirect(url);
 
--- a/src/authkit-callback-route.ts
+++ b/src/authkit-callback-route.ts
@@ -5,56 +5,68 @@ import { WORKOS_CLIENT_ID } from './env-variables.js';
 import { encryptSession } from './session.js';
 import { cookieName, cookieOptions } from './cookie.js';
 
-export async function authkitCallbackRoute(request: NextRequest) {
-  const code = request.nextUrl.searchParams.get('code');
-
-  if (code) {
-    try {
-      // Use the code returned to us by AuthKit and authenticate the user with WorkOS
-      const { accessToken, refreshToken, user, impersonator } = await workos.userManagement.authenticateWithCode({
-        clientId: WORKOS_CLIENT_ID,
-        code,
-      });
-
-      const url = request.nextUrl.clone();
-
-      // Cleanup params
-      url.searchParams.delete('code');
-
-      // Redirect to the requested path and store the session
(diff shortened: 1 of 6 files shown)


src/get-authorization-url.ts
```diff
@@ -1,11 +1,12 @@
 import { workos } from './workos.js';
 import { WORKOS_CLIENT_ID, WORKOS_REDIRECT_URI } from './env-variables.js';
 
-async function getAuthorizationUrl() {
+async function getAuthorizationUrl(returnPathname?: string) {
   return workos.userManagement.getAuthorizationUrl({
     provider: 'authkit',
     clientId: WORKOS_CLIENT_ID,
     redirectUri: WORKOS_REDIRECT_URI,
+    state: returnPathname ? btoa(JSON.stringify({ returnPathname })) : undefined,
```
PaulAsjes, line 9: Clever!

src/authkit-callback-route.ts
```diff
@@ -5,56 +5,68 @@ import { WORKOS_CLIENT_ID } from './env-variables.js';
...
 
-      const url = request.nextUrl.clone();
+        // Cleanup params
+        url.searchParams.delete('code');
+        url.searchParams.delete('state');
 
-      // Cleanup params
-      url.searchParams.delete('code');
+        // Redirect to the requested path and store the session
+        url.pathname = returnPathname ? returnPathname : returnPathnameOption;
```
PaulAsjes, line 35: Could simplify to 
```suggestion
        url.pathname = returnPathname || returnPathnameOption;
```

src/index.ts
```diff
@@ -1,11 +1,11 @@
-import { authkitCallbackRoute } from './authkit-callback-route.js';
+import { handleAuth } from './authkit-callback-route.js';
 import { authkitMiddleware } from './middleware.js';
 import { getUser } from './session.js';
 import { getSignInUrl, signOut } from './auth.js';
 import { Impersonation } from './impersonation.js';
 
 export {
-  authkitCallbackRoute,
+  handleAuth,
```
PaulAsjes, line 8: This kinda breaks backwards compatibility right? Wonder if that means we should release a major version.

PaulAsjes (changes requested):

src/authkit-callback-route.ts
```diff
@@ -5,56 +5,68 @@ import { WORKOS_CLIENT_ID } from './env-variables.js';
...
+  returnPathname?: string;
+}
+
+export function handleAuth(options: HandleAuthOptions = {}) {
+  const { returnPathname: returnPathnameOption = '/' } = options;
+
+  return async function GET(request: NextRequest) {
+    const code = request.nextUrl.searchParams.get('code');
+    const state = request.nextUrl.searchParams.get('state');
+    const returnPathname = state ? JSON.parse(atob(state)).returnPathname : null;
```
mthadley, line 18: What should happen if `state` doesn't contain valid JSON?

Right now it will throw, which is probably fine if the `state` was tampered with (maliciously or not), but wanted to point it out in case anyone had other preferences.

src/index.ts
benoitgrelard: Typically I think that's fine given we're in the 0.x range.

From SemVer:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

src/authkit-callback-route.ts
benoitgrelard: Yeah I think it's fine for now considering it would only be an issue if tampered with.

PaulAsjes (approved):

PR #12, opened 2024-03-27
Rename package

benoitgrelard:


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@workos-inc/nextjs",
+  "name": "@workos-inc/authkit-nextjs",
   "version": "0.3.0",
```
PaulAsjes, line 3: Should probably set the version to 0.1.0 as this'll be published as a whole new package.

PaulAsjes (changes requested):

package.json
benoitgrelard: I'm not sure, it might be nice to explain the continuity from the rename, that the next published version will be 0.4.0. Reason is, there are tags in the codebase for each versions, and I was planning to attach releases to those to show what changed, so on 0.4.0 it'll explain the breaking changes (including package rename).

PaulAsjes: That's fair!

PaulAsjes (approved):

PR #21, opened 2024-03-28
Support automatically pushing releases to NPM

jasonroelofs:
This workflow hooks into the GitHub Release feature, pushing pre-releases to the `next` tag and full releases to the `latest` tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,41 @@
+name: Release
+
+on:
+  # Support manually pushing a new release
+  workflow_dispatch: {}
+  # Trigger when a release is published
+  release:
+    types: [published]
+
+defaults:
+  run:
+    shell: bash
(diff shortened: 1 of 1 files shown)


benoitgrelard (approved):
Thanks for this!

So just so I'm clear how we would use this, this is how I interpret it:
- a version bump would be made in a PR (probably still using `npm version` locally)
- we'd create a new (pre)release in GitHub, and that action would trigger the correct published tag based on if we ticked pre-release or not when creating the release

Is that basically how it works?

jasonroelofs:
Correct! Due to it's nature I'm not directly able to test this workflow until we merge but yeah everything should be handled automatically by using Releases here in the repo.

benoitgrelard:
Cool, I've merged it now, we can try it all next time we have to run a release.

PR #25, opened 2024-04-05
Fix cookies not setting on Safari on localhost

benoitgrelard:
This PR fixes the issue where Safari wouldn't set cookies when used for testing locally.

More info: https://codedamn.com/news/web-development/safari-cookie-is-not-being-set#:~:text=The%20Issue%20with,of%20web%20development.


PaulAsjes (approved):
Great catch!

PR #27, opened 2024-04-05
export `cookieName`, `cookieOptions` and `encryptSession`

nirtamir2:
This PR export `cookieOptions`, `cookieName` and `encryptSession` from WorkOS. 
The motivation to do it is for people who try to implement the authentication themself and need to copy the code from this repository (that may change in the future). Also, I need to install `iron-session` as a dependency in my project if I implement it myself. 

Alternatively, I could create a function that sets the cookie like the SDK instead of writing this code the codebase:

```ts
        const session = await encryptSession({ accessToken, refreshToken, user, impersonator });
        cookies().set(cookieName, session, cookieOptions);
```

### Notes:
I still haven't organized everything properly with my implementation, but it seems like I need to export those parts from the SDK to implement it. Feel free to change or block this PR if you feel it's wrong :)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,8 +1,8 @@
 import { handleAuth } from './authkit-callback-route.js';
 import { authkitMiddleware } from './middleware.js';
-import { getUser } from './session.js';
+import { getUser, encryptSession } from './session.js';
 import { getSignInUrl, getSignUpUrl, signOut } from './auth.js';
-import { cookieName } from './cookie.js';
+import { cookieName, cookieOptions } from './cookie.js';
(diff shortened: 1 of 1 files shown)


PaulAsjes:
Thanks for the PR! We're going to discuss a bit internally on what the best course of action here is. We're thinking in particular on how we can make some high-level abstractions to make this easier on you.

gabrielalmeida:
High-level abstractions would be great but just exporting those would make the library useful to anyone migrating from other session management implementations, right now it assumes you don't have a previous one in place. 

We are doing the same copy/paste + iron-session dependency to properly migrate our users from their previous sessions.

Issue #20, opened 2024-03-28
Allow requiring authenticated user in middleware on a matcher


sirupsen:
Hey! I think it's far too easy to miss a `getUser()` to require authentication in server actions, so we moved all auth to be required in the middleware on URL prefix matches. We probably won't use this library until it's possible to enforce the middleware to auth all paths under some prefix (not just run the logic, as the matcher does), just too risky.

PaulAsjes:
Would you mind elaborating a bit more on what you're after? You can match via a prefix in the middleware, e.g.

```js
export const config = {
  matcher: ['/dashboard/:path*'],
}
```
but your comment suggests that that's not what you're after. Can you talk a bit more about what you'd like to see in this package?

sirupsen:
@PaulAsjes as far as I understand from the docs/skimming the code, this matcher doesn't _force_ authentication on the path, i.e. returning an HTTP 401 if the user isn't authenticated.

It just runs the authentication middlware. The user is responsible for calling `getUser`. This is too insecure for my liking, as it's far too easy to forget calling `getUser` e.g. in a server side mutation.

But maybe I don't have it right? 🤔

PaulAsjes:
Thanks I think I get it now. At the moment we throw an error if you use `getUser` but don't have the path specified in your middleware. You'd like the opposite to be true as well that if you have the path in your middleware but don't call `getUser` we'd also error or return a 401.

I'll talk this over with the team on what we can do for this scenario.

sirupsen:
@PaulAsjes to add concrete facts to the concern, we missed `getUser()` for a server mutation in our application, which was the only thing that showed up on a recent pentest. That's why I created our own middleware (that I'd like to replace with yours) to ensure that we authenticate by default, to not rely on manual calls—like how most auth frameworks I'm used to in Rails, Django, etc. would do 👍🏻

PaulAsjes:
Just so I can get my head around this, what's the use case for wanting a page that is protected by auth but you don't want to fetch the current user?

There's some [discourse](https://twitter.com/chronark_/status/1770721609512321126) in the community on whether your auth should be controlled via middleware or on a per-route basis, hence why I want to make sure I understand the use case before we implement anything.

sirupsen:
[...]

PR #29, opened 2024-04-25
Add auth middleware mode

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/20

Adds a middleware auth mode to the library.

By default auth is invoked by a `getUser` call on a page, however there are some use cases where you wouldn't want this, like if your page doesn't need user data but you still want it protected.

This adds "middleware auth mode", where auth is invoked by the middleware directly instead of on a page-by-page basis. When enabled, middleware auth mode will require sign in for all pages specified in the middleware config matches, unless the page is included in the allowlist for unauthenticated paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Enabling `ensureSignedIn` will redirect users to AuthKit if they attempt to acce
 
 ### Middleware auth
 
-The default behavior of this library is to request authentication via the `getUser` method on a per-page basis. There are some use cases where you don't want to call `getUser` (e.g. you don't need user data for your page) or if you'd prefer a "secure by default" approach where every route defined in your middleware matcher is protected unless specified otherwise. In those cases you an opt to use middleware auth instead:
+The default behavior of this library is to request authentication via the `getUser` method on a per-page basis. There are some use cases where you don't want to call `getUser` (e.g. you don't need user data for your page) or if you'd prefer a "secure by default" approach where every route defined in your middleware matcher is protected unless specified otherwise. In those cases you can opt-in to use middleware auth instead:
 
 ```ts
 import { authkitMiddleware } from '@workos-inc/authkit-nextjs';
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ import { authkitMiddleware } from '@workos-inc/authkit-nextjs';
 export default authkitMiddleware();
 
 // Match against pages that require auth
-// Leave this out if you want auth on every page in your application
-export const config = { matcher: ['/admin'] };
+// Leave this out if you want auth on every resource (including images, css etc.)
+export const config = { matcher: ['/', '/admin'] };
 ```
 
 ## Usage
@@ -122,6 +122,27 @@ const { user } = await getUser({ ensureSignedIn: true });
 
 Enabling `ensureSignedIn` will redirect users to AuthKit if they attempt to access the page without being authenticated.
 
+### Middleware auth
+
+The default behavior of this library is to request authentication via the `getUser` method on a per-page basis. There are some use cases where you don't want to call `getUser` (e.g. you don't need user data for your page) or if you'd prefer a "secure by default" approach where every route defined in your middleware matcher is protected unless specified otherwise. In those cases you an opt to use middleware auth instead:
+
+```ts
+import { authkitMiddleware } from '@workos-inc/authkit-nextjs';
+
+export default authkitMiddleware({
+  middlewareAuth: {
+    enabled: true,
+    unauthenticatedPaths: ['/', '/about'],
+  },
+});
+
+// Match against pages that require auth
+// Leave this out if you want auth on every resource (including images, css etc.)
+export const config = { matcher: ['/', '/admin', '/about'] };
+```
+
+In the above example the `/admin` page will require a user to be signed in, whereas `/` and `/about` can be accessed without signing in.
+
 ### Signing out
 
 Use the `signOut` method to sign out the current logged in user and redirect to your app's homepage. The homepage redirect is set in your WorkOS dashboard settings under "Redirect".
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -122,6 +122,29 @@ const { user } = await getUser({ ensureSignedIn: true });
 
 Enabling `ensureSignedIn` will redirect users to AuthKit if they attempt to access the page without being authenticated.
 
+### Middleware auth
+
+The default behavior of this library is to request authentication via the `getUser` method on a per-page basis. There are some use cases where you don't want to call `getUser` (e.g. you don't need user data for your page) or if you'd prefer a "secure by default" approach where every route defined in your middleware matcher is protected unless specified otherwise. In those cases you an opt to use middleware auth instead:
```
benoitgrelard, line 127:
```suggestion
The default behavior of this library is to request authentication via the `getUser` method on a per-page basis. There are some use cases where you don't want to call `getUser` (e.g. you don't need user data for your page) or if you'd prefer a "secure by default" approach where every route defined in your middleware matcher is protected unless specified otherwise. In those cases you can opt-in to use middleware auth instead:
```

src/session.ts
```diff
@@ -32,14 +35,27 @@ async function updateSession(request: NextRequest, debug: boolean) {
...
+
   // If no session, just continue
   if (!session) {
     return NextResponse.next({
       request: { headers: newRequestHeaders },
     });
   }
 
-  const hasValidSession = await verifyAccessToken(session.accessToken);
+  const hasValidSession: boolean = await verifyAccessToken(session.accessToken);
```
benoitgrelard, line 58: Why did we need to add a type annotation here? Isn't this inferred anyway?

src/session.ts
```diff
@@ -87,6 +103,22 @@ async function updateSession(request: NextRequest, debug: boolean) {
   }
 }
 
+function parseRoute(route: string): string {
```
benoitgrelard, line 106: I find the naming confusing here, it's not really parsing a route, it's taking a global-like path expression and returning a regex string to test if a path matches it.

Could we rename it to something more meaningful/specific and also make it return the regexp itself:

```ts
function getMiddlewareAuthPathRegex(pathGlob: string) {
  …
  regex = new RegExp(tokensToRegexp(tokens).source);
  …
}
```

src/session.ts
```diff
@@ -7,7 +7,10 @@ import { cookieName, cookieOptions } from './cookie.js';
 import { workos } from './workos.js';
 import { WORKOS_CLIENT_ID, WORKOS_COOKIE_PASSWORD } from './env-variables.js';
 import { getAuthorizationUrl } from './get-authorization-url.js';
-import { AccessToken, NoUserInfo, Session, UserInfo } from './interfaces.js';
+import { AccessToken, AuthkitMiddlewareAuth, NoUserInfo, Session, UserInfo } from './interfaces.js';
+
+import { parse, tokensToRegexp } from 'path-to-regexp';
+import { parse as parseURL } from 'url';
```
benoitgrelard, line 13: Where does this dependency come from? It looks like we're relying on a transitive dep at the moment as we haven't added it to the package.json?

src/interfaces.ts
```diff
@@ -40,3 +40,13 @@ export interface GetAuthURLOptions {
   screenHint?: 'sign-up' | 'sign-in';
   returnPathname?: string;
 }
+
+export interface AuthkitMiddlewareAuth {
+  enabled: boolean;
```
benoitgrelard, line 45: do we need this extra `enabled` key?

src/session.ts
PaulAsjes: Oops, didn't mean to commit this.

src/session.ts
PaulAsjes: It's part of the Node core library, but turns out it's actually deprecated so I'll refactor this.

src/interfaces.ts
PaulAsjes: Unfortunately I believe we do. Having `unauthenticatedPaths` isn't enough to determine whether a user wants to use middleware auth or not. 

You could argue that if the `authenticatedPaths` array is empty then that means that the user doesn't want to use middleware auth. However an edge case is that someone wants middleware auth on _all_ routes with no exception, in which case an empty array makes sense.

Having an explicit `enabled` boolean means there can be no confusion on what the user's intention is.

benoitgrelard: I see, I guess it doesn't hurt to be extra explicit.

src/session.ts
```diff
@@ -32,6 +34,18 @@ async function updateSession(request: NextRequest, debug: boolean) {
...
+    const routeRegex: RegExp = getMiddlewareAuthPathRegex(route);
+
+    return routeRegex.exec(request.nextUrl.pathname);
+  });
+
+  // If the user is logged out and this path isn't on the allowlist for logged out paths, redirect to AuthKit.
+  if (middlewareAuth.enabled && matchedPaths.length === 0 && !session) {
+    if (debug) console.log('Unauthenticated user on protected route, redirecting to AuthKit');
+    return NextResponse.redirect(await getAuthorizationUrl({ returnPathname: new URL(request.url).pathname }));
+  }
```
benoitgrelard, line 47: nit:

```suggestion
  const matchedPaths: string[] = middlewareAuth.unauthenticatedPaths.filter((pathGlob) => {
    const pathRegex = getMiddlewareAuthPathRegex(pathGlob);

    return pathRegex.exec(request.nextUrl.pathname);
  });

  // If the user is logged out and this path isn't on the allowlist for logged out paths, redirect to AuthKit.
  if (middlewareAuth.enabled && matchedPaths.length === 0 && !session) {
    if (debug) console.log('Unauthenticated user on protected route, redirecting to AuthKit');
    return NextResponse.redirect(await getAuthorizationUrl({ returnPathname: new URL(request.url).pathname }));
  }
```

src/session.ts
```diff
@@ -87,6 +101,23 @@ async function updateSession(request: NextRequest, debug: boolean) {
...
+
+    const tokens = parse(parsedPath);
+    regex = tokensToRegexp(tokens).source;
+
+    return new RegExp(regex);
+  } catch (err) {
+    throw new Error(`Error parsing routes for middleware auth. Reason: ${err.message}`);
+  }
+}
+
```
benoitgrelard, line 120: nit:
```suggestion
function getMiddlewareAuthPathRegex(pathGlob: string) {
  let regex: string;

  try {
    // Redirect URI is only used to construct the URL
    const url = new URL(pathGlob, WORKOS_REDIRECT_URI);
    const path = `${url.pathname!}${url.hash || ''}`;

    const tokens = parse(path);
    regex = tokensToRegexp(tokens).source;

    return new RegExp(regex);
  } catch (err) {
    throw new Error(`Error parsing routes for middleware auth. Reason: ${err.message}`);
  }
}

```

benoitgrelard (commented):
A few nits and we're good I think.

benoitgrelard (approved):

PR #43, opened 2024-05-10
add `getAccessToken`

### LINKED ISSUE #42: Is there a way to access the user token?
I was using clerk for authentication with supabase but now I'm trying to switch to work os. Is there a way to access the user's token to work with supabase?
```
import { createClient } from "@supabase/supabase-js";
import { auth } from "@clerk/nextjs";

export default async function supabaseClient() {
  const { userId, getToken } = auth();
  if (!userId) throw new Error("User not authenticated");

  const accessToken = await getToken({ template: "supabase" });

  const supabase = createClient(
    process.env.NEXT_PUBLIC_SUPABASE_URL ?? "",
    process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY ?? "",
    {
      global: { headers: { Authorization: `Bearer ${accessToken}` } },
    }
  );
  return supabase;
}
```

cmatheson:
fixes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,30 @@ export default function App() {
 }
 ```
 
+### Get the access token
+
+Sometimes it is useful to obtain the access token directly (e.g. to make API requests to another service).
+
+```jsx
+import { getAccessToken } from '@workos-inc/authkit-nextjs';
+
+export default async function CoolComponent() {
+  const accessToken = await getAccessToken();
+
+  if (!accessToken) {
+    return <div>Not signed in</div>;
+  }
+
+  const serviceData = await fetch('/api/path', {
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+    },
+  });
+
+  return <div>{serviceData}</div>;
+}
+```
+
 ### Debugging
 
 To enable debug logs, initialize the middleware with the debug flag enabled.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -167,6 +167,30 @@ export default function App() {
...
+
+Sometimes it is useful to obtain the access token directly (e.g. to make API requests to another service).
+
+```jsx
+import { getAccessToken } from '@workos-inc/authkit-nextjs';
+
+export default async function CoolComponent() {
+  const accessToken = await getAccessToken();
+
+  if (!accessToken) {
```
cmatheson, line 180: it's kind of annoying to have to wonder if you're getting an access token back or not. i wonder if `accessToken` should instead (or additionally?) be another piece of data returned from `getUser` (that way you choose the `ensureSignedIn` behavior).

PaulAsjes: I'm leaning towards `accessToken` being a part of the data returned from `getUser`. 

Eventually we might want to rename `getUser` to something like `getAuth` to make it clearer you can do this:

```js
const { accessToken, user } = await getAuth();
```

README.md
```diff
@@ -167,6 +167,30 @@ export default function App() {
 }
 ```
 
+### Get the access token
+
+Sometimes it is useful to obtain the access token directly (e.g. to make API requests to another service).
```
PaulAsjes, line 172: This is a little vague, could use a better example on what you'd do with the access token.

PaulAsjes (approved):

PR #48, opened 2024-05-30
Set max age on cookie

PaulAsjes:
Sessions were not persisting on Chrome or iOS when the browser was completely shut down. This fixes that by adding a `maxAge` value to the cookie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ To use the `signOut` method, you'll need to set your app's homepage in your Work
 Certain environment variables are optional and can be used to debug or configure cookie settings.
 
 ```sh
-WORKOS_COOKIE_MAX_AGE='600' # maximum age of the cookie in seconds. Defaults to 10 minutes
+WORKOS_COOKIE_MAX_AGE='600' # maximum age of the cookie in seconds. Defaults to 31 days
 WORKOS_API_HOSTNAME='api.workos.com' # base WorkOS API URL
 WORKOS_API_HTTPS=true # whether to use HTTPS in API calls
 WORKOS_API_PORT=3000 # port to use for API calls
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -35,6 +35,17 @@ openssl rand -base64 24
 
 To use the `signOut` method, you'll need to set your app's homepage in your WorkOS dashboard settings under "Redirects".
 
+### Optional configuration
+
+Certain environment variables are optional and can be used to debug or configure cookie settings.
+
+```sh
+WORKOS_COOKIE_MAX_AGE='600' # maximum age of the cookie in seconds. Defaults to 10 minutes
```
cmatheson, line 43: i think we should make the max age quite long (imagine you close your browser (or your OS closes it for you) and you don't come back to it for several hours/days).

since the session is being verified separately from the cookie anyway, i think we should set this to something like 400 days (which is [the maximum allowed by chrome](https://developer.chrome.com/blog/cookie-max-age-expires)).

but if we feel like that is too long, at least a month would avoid any noticeable logouts.

PaulAsjes: How about 31 days as the default?

cmatheson: 31 days is OK for most apps I think. If we maximize the cookie age, I think we can make this work for all apps (I don't see a downside to making the cookie last "forever" the refresh_token/access_token are the actual time-limited aspects of the session.

(No one should have to think about this setting, but they will have to when it's too short).

cmatheson (approved):
(i'm in favor of going longer on the cookie, but either way this is way better than what we have right now).

README.md
PaulAsjes: 👍 you've convinced me, changing it to default to the max allowed number of days.

Issue #40, opened 2024-05-05
MiddlewareAuth enabled blocks auth/callback by default


ryendu:
Hey, I setup middlewareAuth as introduced in #29 to prevent unauthorized users from accessing most endpoints. However, I kept running into a bug where unauthenticated users attempting to sign in would just keep getting redirected to the sign in page after attempting to sign in if `middlewareAuth` is enabled. I did some debugging and realized that by default if `middlewareAuth` is enabled, the `/auth/callback` endpoint that handles user sign in in is also blocked and users are prevented from signing in. The easy fix would to be to add `'/auth/callback'` or one's callback route to the `unauthenticatedPaths`, but I wanted to bring this up because I thought this should be made clear in the documentation or somewhere to prevent others from running into this as well and getting confused.

PaulAsjes:
Thanks for the report, you likely had `/auth/callback` in your middleware matcher which would trigger the middleware after login. In that case if you don't have the callback URL added to `unauthenticatedPaths` the "check if user is auth'd" logic hits before the session logic can execute, putting you in a login loop.

An easier fix would be to just leave `/auth/callback` out of your middleware matcher, but I'll make sure to update the docs to make this clearer.

PR #49, opened 2024-05-31
Throw error on URI mismatch

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/40

In the case where you have:
* Middleware auth mode enabled
* Redirect URL in the middleware matcher
* Don't have the redirect URL in the `unauthenticatedPaths` array

You'd get stuck in a login loop as you'd be redirected to AuthKit before the session can be set establishing that you've already logged in.

This adds a check and throws an error in that case.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -40,6 +40,18 @@ async function updateSession(request: NextRequest, debug: boolean, middlewareAut
     return pathRegex.exec(request.nextUrl.pathname);
   });
 
+  const url = new URL(WORKOS_REDIRECT_URI);
+
+  if (
+    middlewareAuth.enabled &&
+    url.pathname === request.nextUrl.pathname &&
+    !middlewareAuth.unauthenticatedPaths.includes(url.pathname)
+  ) {
+    throw new Error(
+      `Mismatch detected: WorkOS redirect URI '${url.pathname}' found in middleware matcher but is not included in 'unauthenticatedPaths' array. This will cause a log in loop. Add '${url.pathname}' to 'unauthenticatedPaths' config to resolve.`,
+    );
+  }
+
   // If the user is logged out and this path isn't on the allowlist for logged out paths, redirect to AuthKit.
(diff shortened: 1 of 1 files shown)


src/session.ts
```diff
@@ -40,6 +40,18 @@ async function updateSession(request: NextRequest, debug: boolean, middlewareAut
...
   });
 
+  const url = new URL(WORKOS_REDIRECT_URI);
+
+  if (
+    middlewareAuth.enabled &&
+    url.pathname === request.nextUrl.pathname &&
+    !middlewareAuth.unauthenticatedPaths.includes(url.pathname)
+  ) {
+    throw new Error(
```
cmatheson, line 50: instead of throwing an error, i think we should just push the REDIRECT_URL onto unauthenticatedPaths (could probably be done at startup instead of in the middleware).

src/session.ts
```diff
@@ -34,6 +34,23 @@ async function updateSession(request: NextRequest, debug: boolean, middlewareAut
...
+    !middlewareAuth.unauthenticatedPaths.includes(url.pathname)
+  ) {
+    // In the case where:
+    // - We're using middleware auth mode
+    // - The redirect URI is in the middleware matcher
+    // - The redirect URI isn't in the unauthenticatedPaths array
+    //
+    // then we would get stuck in a login loop due to the redirect happening before the session is set.
+    // It's likely that the user accidentally forgot to add the path to unauthenticatedPaths, so we add it here.
+    middlewareAuth.unauthenticatedPaths.push(url.pathname);
```
cmatheson, line 51: :tada:

 tiny nit: i think this could be done in `middleware.ts` on line 9 (prior to the return) instead of on every request.

cmatheson (approved):

src/session.ts
PaulAsjes: Unfortunately I don't think we can because we don't have access to `request.nextUrl` in that scope.

PR #53, opened 2024-06-26
Add support for fetching permissions JWT claim

mattgd:
Add support for fetching `permissions` JWT claim for RBAC.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -152,13 +152,14 @@ async function getUser({ ensureSignedIn = false } = {}) {
     return { user: null };
   }
 
-  const { sid: sessionId, org_id: organizationId, role } = decodeJwt<AccessToken>(session.accessToken);
+  const { sid: sessionId, org_id: organizationId, role, permissions } = decodeJwt<AccessToken>(session.accessToken);
 
   return {
     sessionId,
     user: session.user,
     organizationId,
     role,
+    permissions,
     impersonator: session.impersonator,
     accessToken: session.accessToken,
   };
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,15 +1,16 @@
-import { redirect } from 'next/navigation';
-import { cookies, headers } from 'next/headers';
-import { NextRequest, NextResponse } from 'next/server';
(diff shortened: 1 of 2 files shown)


src/session.ts
```diff
@@ -1,13 +1,13 @@
...
+import { redirect } from 'next/navigation';
 import { NextRequest, NextResponse } from 'next/server';
-import { jwtVerify, createRemoteJWKSet, decodeJwt } from 'jose';
-import { sealData, unsealData } from 'iron-session';
 import { cookieName, cookieOptions } from './cookie.js';
-import { workos } from './workos.js';
 import { WORKOS_CLIENT_ID, WORKOS_COOKIE_PASSWORD, WORKOS_REDIRECT_URI } from './env-variables.js';
 import { getAuthorizationUrl } from './get-authorization-url.js';
 import { AccessToken, AuthkitMiddlewareAuth, NoUserInfo, Session, UserInfo } from './interfaces.js';
+import { workos } from './workos.js';
```
PaulAsjes, line 10: Why the re-org of imports?

src/session.ts
```diff
@@ -152,15 +152,29 @@ async function getUser({ ensureSignedIn = false } = {}) {
     return { user: null };
   }
 
-  const { sid: sessionId, org_id: organizationId, role } = decodeJwt<AccessToken>(session.accessToken);
+  const { sid: sessionId, org_id: organizationId, role, permissions } = decodeJwt<AccessToken>(session.accessToken);
+
+  const hasPermission = (permission: string) => {
```
PaulAsjes, line 157: I don't think this is necessary since it's a pretty straightforward operation and just complicates the return object. Since `permissions` exists and is included in the return object it's trivial for developers to determine whether the user has a certain permission themselves.

PaulAsjes (changes requested):

src/session.ts
mattgd: Sounds good, will remove!

src/session.ts
mattgd: Just what the `prettier` doctor order it seems. Can revert 👍

PaulAsjes (approved):

Issue #58, opened 2024-06-27
Allow the customization of cookieName


mattbf:
### Issue
If you are using workOS to integrate into a full stack app (a separate backend and frontend) then you have to set the cookie in the backend and consume it in the frontend. 

The AuthKit package for nextjs will only [consume the cookie](https://github.com/workos/authkit-nextjs/blob/8e54010895234d422688ddaf2599191f6b54dfb4/src/session.ts#L184) properly if you set the cookieName in the backend to "wos-session". This is because it's hardcoded [here](https://github.com/workos/authkit-nextjs/blob/8e54010895234d422688ddaf2599191f6b54dfb4/src/cookie.ts#L6) and then consumed elsewhere as cookieName.

PaulAsjes:
Fixed in v0.11.0.

Issue #45, opened 2024-05-22
env vars loaded on require / import


robcresswell:
https://github.com/workos/authkit-nextjs/blob/3eaea54ac89648a70077128b3e9c9dc96c7e2e5c/src/env-variables.ts

I was curious if there was any particular reason for the loading of env vars on require, rather than via some kind of load function. I'm not super familiar with next, but this seems fairly painful because it means you need your environment to be correct just to build the code, let alone run it?

PaulAsjes:
Hey @robcresswell, can you elaborate on why it's painful for you? The library needs all those values to function, so we thought it would be better to fail fast if any are missing.

robcresswell:
Hey @PaulAsjes, thanks for the quick response.

Its mostly around testing. Things like build steps and simple smoke tests ("is the server up and replying to healthchecks") can't be carried out without needing to run in an environment that has all of our WorkOS configuration. There are workarounds for this like dynamic imports AFAIK. It may also be an "us" problem; as I said, not super familiar with next so it might just be my bad usage.

PaulAsjes:
Thanks for elaborating! The library only checks if the env variables are set, not if they are correct or not. Meaning you can put any string value in there and it'll work, with the exception of `WORKOS_COOKIE_PASSWORD` which does need to be 32 characters long.

I think your situation might be a bit of an edge case, which shouldn't be too hard to circumvent. When starting your dev server for testing you can pass in the variables via the command line to make things easier for you when testing, e.g.

```bash
WORKOS_API_KEY=foo WORKOS_CLIENT_ID=bar WORKOS_REDIRECT_URI=baz WORKOS_COOKIE_PASSWORD=longstringthatisatleast32characterslong next dev
```

Does that help?

robcresswell:
Sure; we're aware there are workarounds, I just thought it was worth flagging that this might cause issues. If its not really seen as a problem, I think we can close this issue 😄

robcresswell:
If anyone else stumbles across this, we've ended up with the following in one of our containers:

```
RUN env \
    WORKOS_CLIENT_ID=must-be-set-to-lint \
    WORKOS_API_KEY=must-be-set-to-lint \
    WORKOS_REDIRECT_URI=data:must-be-set-to-lint \
    WORKOS_COOKIE_PASSWORD=must-be-set-to-lint-and-must-be-really-long \
    npm run build
```

Still not a big fan of this behaviour, but it is what it is cc @PaulAsjes

[...]

Issue #24, opened 2024-04-04
Wildcards in redirect URI do not work as expected


tschumwalleye:
I have branch URLs from Vercel preview deployments. 

My `WORKOS_REDIRECT_URI` is configured as a wildcard in my dashboard following the guidelines here: [WorkOS Docs](https://workos.com/docs/sso/redirect-uris/wildcard-characters).

After migrating to authkit-nextjs, redirected users land on a url where the `*` is replaced with `%2A`. 

https://*.example.com -> https://%2A.example.com

I would be able to pass in the URL at run-time if that was an option, which is how I handled this behavior before migrating. I would take advantage of the `VERCEL_URL` environment variable when calling the `getAuthorizationUrl` method.

PaulAsjes:
Thanks for reporting this, that's definitely unintentional. We'll look into this and get back to you!

benoitgrelard:
@tschumwalleye I'd like to ask a few questions to make sure I understand the issue:
- You have added `https://*.example.com` as a redirect URI in the dashboard?
-  What is the value of `WORKOS_REDIRECT_URI` that you pass in? Is it the same URL?

I feel like the issue is that you are unable to pass in the actual URL (one that would match the wildcard scheme) at runtime. Is that correct?

If so, I think we just need to add an option to the `getSignInUrl` so it can be passed at runtime.

benoitgrelard:
I think `signOut` is different though, the URL it redirects to is configured in the dashboard.

<img width="606" alt="CleanShot 2024-04-05 at 14 18 15@2x" src="https://github.com/workos/authkit-nextjs/assets/1539897/f1941505-62d8-45c8-9fba-5a6288eb552f">

tschumwalleye:
That is correct, this leaves me with the same issue of being unable to pass in an accurate URL for preview deployments. I can bypass using the signOut function by calling `cookies().delete("wos-session")` myself, then redirecting. My concerns with this are:
- If that string changes in the future
- Are idp initiated SSO users redirected to a different logout page? (Which would be provided by `getLogoutUrl`)

jxdp:
[vercel-system-vars]: https://vercel.com/docs/projects/environment-variables/system-environment-variables#framework-environment-variables

@tschumwalleye I found a way to work around this using the [experimental instrumentation hook](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation).

It isn't elegant, but it works...

```ts
// instrumentation.ts
export function register() {
[...]

PR #59, opened 2024-06-27
added a configurable WORKOS_COOKIE_NAME optional env var

mattbf:
Fixes issue #58 

Right now, if you set a cookie on your backend and try to use AuthKit with nextjs on your frontend, it will not work unless you set the cookieName to "wos-session".

This would allow users who are integrating with workOS in a full stack application to customize the cookieName used. For example, in a nodejs backend: 
```ts
res.cookie("myapp-session", encryptedSession, {
  path: "/",
  httpOnly: true,
  secure: true,
  sameSite: "lax",
});
```

Then you would set the optional env var `WORKOS_COOKIE_NAME` on the frontend and AuthKit would automatically and properly consume the cookie 🍪 😄


PaulAsjes:
Thanks for this! The code looks good but I'm going to hold on merging this till we figure out a solution for https://github.com/workos/authkit-nextjs/issues/24 and https://github.com/workos/authkit-nextjs/issues/45.

src/session.ts
```diff
@@ -5,14 +5,14 @@ import { jwtVerify, createRemoteJWKSet, decodeJwt } from 'jose';
...
+import { WORKOS_CLIENT_ID, WORKOS_COOKIE_PASSWORD, WORKOS_REDIRECT_URI, WORKOS_COOKIE_NAME } from './env-variables.js';
 import { getAuthorizationUrl } from './get-authorization-url.js';
 import { AccessToken, AuthkitMiddlewareAuth, NoUserInfo, Session, UserInfo } from './interfaces.js';
 
 import { parse, tokensToRegexp } from 'path-to-regexp';
 
-const sessionHeaderName = 'x-workos-session';
-const middlewareHeaderName = 'x-workos-middleware';
+const sessionHeaderName = !!WORKOS_COOKIE_NAME ? `x-${WORKOS_COOKIE_NAME}` : 'x-workos-session';
+const middlewareHeaderName = !!WORKOS_COOKIE_NAME ? `x-${WORKOS_COOKIE_NAME}-middleware` : 'x-workos-middleware';
```
PaulAsjes, line 15: I don't think these need to be configurable. These are specifically used for some behind-the-scenes work (mostly to improve the DX of working with Next.js middleware) and aren't intended for use outside of this library.

PaulAsjes (changes requested):

src/session.ts
mattbf: Roger! I removed these edits

PaulAsjes:
Released a variation of this in 0.11.0.

mattbf:
Cheers, thanks @PaulAsjes!

PR #61, opened 2024-06-28
Release v1

PaulAsjes:
Bumps the version of workos-node, move it to peer dependencies and officially releases as v1 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,14 @@
 {
   "name": "@workos-inc/authkit-nextjs",
-  "version": "0.6.1",
+  "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


jxdp:
Sorry to be that guy, but can you please take a look at https://github.com/workos/authkit-nextjs/issues/24 again. The current API has some severe limitations for anyone deploying on Vercel (probably a lot of your potential customers) and I think it would be highly beneficial to resolve those before releasing V1.

PaulAsjes:
Thanks @jxdp, I agree with you that we should fix that before releasing a v1. I'll put this on hold for now while we address the other issue.

PR #63, opened 2024-07-08
update user when refreshing

cmatheson:
Previously we were only updating the tokens in the session. Now we update the user information too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "@workos-inc/node": "^6.8.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "@workos-inc/node": "^6.8.0",
+    "@workos-inc/node": "^7.12.0",
```
PaulAsjes, line 24: Since we're updating to v7, we should add the `appInfo` object when initializing the WorkOS client. That can go in the version bump PR though.

PaulAsjes (approved):

PR #64, opened 2024-07-09
v0.7.0

cmatheson:
* Updates user information when refreshing
* Bumps WorkOS client to 7.12.0

### DIFF
--- a/src/workos.ts
+++ b/src/workos.ts
@@ -1,11 +1,16 @@
 import { WorkOS } from '@workos-inc/node';
 import { WORKOS_API_HOSTNAME, WORKOS_API_HTTPS, WORKOS_API_KEY, WORKOS_API_PORT } from './env-variables.js';
 
+export const VERSION = '0.7.0';
 
 const options = {
   apiHostname: WORKOS_API_HOSTNAME,
   https: WORKOS_API_HTTPS ? WORKOS_API_HTTPS === 'true' : true,
   port: WORKOS_API_PORT ? parseInt(WORKOS_API_PORT) : undefined,
+  appInfo: {
+    name: 'authkit/nextjs',
+    version: VERSION,
+  },
 };
 
 // Initialize the WorkOS client
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@workos-inc/authkit-nextjs",
-  "version": "0.6.1",
+  "version": "0.7.0",
   "lockfileVersion": 3,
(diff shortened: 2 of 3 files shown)


src/workos.ts
```diff
@@ -1,11 +1,16 @@
 import { WorkOS } from '@workos-inc/node';
 import { WORKOS_API_HOSTNAME, WORKOS_API_HTTPS, WORKOS_API_KEY, WORKOS_API_PORT } from './env-variables.js';
+import { version } from '../package.json';
 
 
 const options = {
   apiHostname: WORKOS_API_HOSTNAME,
   https: WORKOS_API_HTTPS ? WORKOS_API_HTTPS === 'true' : true,
   port: WORKOS_API_PORT ? parseInt(WORKOS_API_PORT) : undefined,
+  appInfo: {
```
cmatheson, line 10: this is new in WorkOS v7

tsconfig-base.json
```diff
@@ -5,6 +5,7 @@
     "jsx": "react",
     "module": "ES2020",
     "moduleResolution": "Node",
+    "resolveJsonModule": true,
```
mthadley, line 8: Nice. Any reason not to turn this on and do the same in `@workos-inc/node` for the [`VERSION` constant](https://github.com/workos/workos-node/blob/80edc61076eb9a2749756051a083a947198497cf/src/workos.ts#L34)?

mthadley (approved):

src/workos.ts
```diff
@@ -1,11 +1,16 @@
 import { WorkOS } from '@workos-inc/node';
 import { WORKOS_API_HOSTNAME, WORKOS_API_HTTPS, WORKOS_API_KEY, WORKOS_API_PORT } from './env-variables.js';
+import { version } from '../package.json';
```
PaulAsjes, line 3: I'm fairly certain this will break in production, as our tsconfig doesn't know how to handle json files.

PaulAsjes (changes requested):

src/workos.ts
PaulAsjes: I'm dumb and didn't see that you added `resolveJsonModule`.

PaulAsjes (approved):

src/workos.ts
cmatheson: well, the json part is fine but it actually does break production 😂 😭 . it changes the compiled output such that all the ts files get nested under `src` and breaks things. going to push up a new iteration.

PR #67, opened 2024-07-22
Preserve search params on return

PaulAsjes:
Search params were not being saved on return from AuthKit, this fixes that.

### DIFF
--- a/src/authkit-callback-route.ts
+++ b/src/authkit-callback-route.ts
@@ -33,12 +33,12 @@ export function handleAuth(options: HandleAuthOptions = {}) {
 
         // Extract the search params if they are present
         if (returnPathname.includes('?')) {
-          const [pathname, search] = returnPathname.split('?');
-          url.pathname = pathname;
-          search.split('&').forEach((param: string) => {
-            const [key, value] = param.split('=');
+          const newUrl = new URL(returnPathname, 'https://example.com');
+          url.pathname = newUrl.pathname;
+
+          for (const [key, value] of newUrl.searchParams) {
             url.searchParams.append(key, value);
-          });
+          }
         } else {
           url.pathname = returnPathname;
         }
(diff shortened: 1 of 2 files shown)


src/authkit-callback-route.ts
```diff
@@ -29,7 +29,19 @@ export function handleAuth(options: HandleAuthOptions = {}) {
         url.searchParams.delete('state');
 
         // Redirect to the requested path and store the session
-        url.pathname = returnPathname ?? returnPathnameOption;
+        returnPathname = returnPathname ?? returnPathnameOption;
+
+        // Extract the search params if they are present
+        if (returnPathname.includes('?')) {
```
cmatheson, line 35: instead of manually splitting the params out of the url, can we access all the existing query params using searchParams too? Something like:

```!js
for (const [key, value] of new URL(returnPathname, "https://example.com").searchParams) {
  url.searchParams.append(key, value)
}
```

src/authkit-callback-route.ts
```diff
@@ -29,7 +29,19 @@ export function handleAuth(options: HandleAuthOptions = {}) {
...
+          const newUrl = new URL(returnPathname, 'https://example.com');
+          url.pathname = newUrl.pathname;
+
+          for (const [key, value] of newUrl.searchParams) {
+            url.searchParams.append(key, value);
```
cmatheson, line 40: have you tested this? normally `url.searchParams` is [readonly](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams). (going to approve assuming you have verified this).

cmatheson (approved):

src/authkit-callback-route.ts
cmatheson: (i think i'm just reading the docs wrong, this looks good :+1:)

PaulAsjes: Yup works fine in my testing. Also works when goofing around in the chrome console:

CleanShot 2024-07-22 at 20 37 55

`url.searchParams` is read-only, meaning `url.searchParams = 'foo'` does nothing. However it has a valid `append` method to add stuff to it.

PR #68, opened 2024-07-22
Add warning for CORS issue 

PaulAsjes:
When using middleware auth mode and Next.js server actions, it's possible to trigger a CORS issue. The edge case manifests itself when:

1. User doesn't have a valid session (i.e. logged out in a separate tab)
2. User triggers a server action in the now stale page
3. CORS error surfaces in the browser console

The above happens because without a valid session and in middleware auth mode, the POST request to the server action is intercepted and a redirect to AuthKit is attempted, which triggers a CORS error. 

The solution here is to wrap the server action call in a `try/catch` block so an error can displayed or the user redirected client side.

Also updates Next.js to its latest minor version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ In the above example the `/admin` page will require a user to be signed in, wher
 
 ### Middleware auth and server actions
 
-Using [server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) with middleware auth mode can result in unexpected CORS errors if you don't wrap your server action call in a try/catch block. This is because a call to a server action without an existing session will be intercepted by the middleware and a redirect to AuthKit will be attempted, which will trigger a CORS error.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -156,6 +156,42 @@ In the above example the `/admin` page will require a user to be signed in, wher
 
 `unauthenticatedPaths` uses the same glob logic as the [Next.js matcher](https://nextjs.org/docs/pages/building-your-application/routing/middleware#matcher).
 
+### Middleware auth and server actions
+
+Using [server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) with middleware auth mode can result in unexpected CORS errors if you don't wrap your server action call in a try/catch block. This is because a call to a server action without an existing session will be intercepted by the middleware and a redirect to AuthKit will be attempted, which will trigger a CORS error.
```
southgate, line 161: > which will trigger a CORS error.

triggering a CORS error.

PR #73, opened 2024-08-02
Add support for `organizationId` parameter when signing in or refreshing session

mthadley:
Makes the following changes to the public interface:

* Adds a new `organizationId` parameter to `getSignInUrl`
* Adds a new `refreshSession` helper function that also accepts an `organizationId` parameter`

Internally, it the middleware is updated to pass the current `org_id` from the `access_token` when automatically refreshing sessions, in order to preserve the currently authorized organization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,12 @@ export default async function HomePage() {
 }
 ```
 
+### Refreshing the session
+
+Use the `refreshSession` method in a server action or route handler to fetch the latest session details, including any changes to the user's roles or permissions.
+
+The `organizationId` parameter can be passed to `refreshSession` in order to "switch" to switch the session to a different organization. If the current session is not authorized for the next organization, an appropriate [authentication error](https://workos.com/docs/reference/user-management/authentication-errors) will be returned.
+
 ### Debugging
 
 To enable debug logs, initialize the middleware with the debug flag enabled.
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "@workos-inc/node": "7.14.0",
+    "@workos-inc/node": "7.18.0",
```
PaulAsjes, line 24: Tiny nit: we're on `7.20.0` as of today.

PaulAsjes (changes requested):
Can we add some instructions to the readme on how and when to use `refreshSession`?

README.md
```diff
@@ -208,6 +208,12 @@ export default async function HomePage() {
 }
 ```
 
+### Refreshing the session
+
+Use the `refreshSession` method in a server action or route handler to fetch the latest session details, including any changes to the user's roles or permissions.
+
+The `organizationId` parameter can be passed to `refreshSession` in order to "switch" to switch the session to a different organization. If the current session is not authorized for the next organization, an appropriate [authentication error](https://workos.com/docs/reference/user-management/authentication-errors) will be returned.
```
PaulAsjes, line 215:
```suggestion
The `organizationId` parameter can be passed to `refreshSession` in order to switch the session to a different organization. If the current session is not authorized for the next organization, an appropriate [authentication error](https://workos.com/docs/reference/user-management/authentication-errors) will be returned.
```

PaulAsjes (approved):
Tiny wordsmithing nit, otherwise good!

PR #74, opened 2024-08-02
Allow setting cookie `Domain` attribute

cmatheson:
This setting would be useful when you want to share the workos session between two different apps/subdomains. (NOTE: the COOKIE_PASSWORD would need to be the same on both apps).


README.md
```diff
@@ -50,6 +50,9 @@ WORKOS_COOKIE_MAX_AGE='600' # maximum age of the cookie in seconds. Defaults to
 WORKOS_API_HOSTNAME='api.workos.com' # base WorkOS API URL
 WORKOS_API_HTTPS=true # whether to use HTTPS in API calls
 WORKOS_API_PORT=3000 # port to use for API calls
+# the domain (and subdomains) that the session cookie will be available on.
+# (normally not needed).
```
PaulAsjes, line 54:
```suggestion
# the domain (and subdomains) that the session cookie will be available on
```

README.md
```diff
@@ -50,6 +50,9 @@ WORKOS_COOKIE_MAX_AGE='600' # maximum age of the cookie in seconds. Defaults to
...
+WORKOS_COOKIE_DOMAIN='example.com'
 ```
```
PaulAsjes, line 57:
```suggestion
`WORKOS_COOKIE_DOMAIN` can be used to share WorkOS sessions between apps/domains. 

Note: The `WORKOS_COOKIE_PASSWORD` would need to be same across apps/domains.
```

Not needed for most use cases

PaulAsjes (approved):
Just a few docs suggestions, looks good otherwise.

PR #75, opened 2024-08-06
Check for valid session when tab becomes active

lucasmotta:
## Description
Add a new `Provider` component that users should wrap their application in order to check for invalid sessions, in case for example, a tab stayed open for too long or the user signed out in a different tab.

If the user has a invalid session upon returning to a tab, the page will be reloaded so the middleware can take care of the redirection.

By forcing this page reload, we prevent [possible CORS issues](https://github.com/workos/authkit-nextjs/pull/68) as server actions would throw an error as you are not authenticated.

The current use of the `Provider` is to listen to tab events and check for invalid session, but we can eventually introduce new feature inside the Provider, like passing down the current user in a context so it can be accessed in client components.

## How does that work?

When the user switches back to the app, we make a call to a server action, which in Next.js world is just a `POST` request to the current route the user is in.

If the user has a valid session, the request will most likely be successful so we don't need to do anything.
But if the user does not have an active session, there are two outcomes:
- The user is in a protected page, the server action will fail because the of the middleware, so we can reload the page to let the middleware handle the redirection
- If the user is a `public` page, the server action will succeed and don't have to do anything

And also, if the user does not have the `middlewareAuth` enabled, the server action will also be successful, so we don't reload the page.

## Open questions
- Should this behaviour be configured via a prop, like `<Provider checkForInactiveTab />` ?
- Should we allow users to customize what to do in the case of a invalid session? For example instead of reloading, display a React component saying that they are logged out?

### DIFF
--- a/src/provider.tsx
+++ b/src/provider.tsx
@@ -1,0 +1,29 @@
+'use client';
+
+import * as React from 'react';
+import { checkSessionAction } from './actions.js';
+
+export const Provider = ({ children }: React.PropsWithChildren) => {
+  React.useLayoutEffect(() => {
+    const handleVisibilityChange = async () => {
+      if (document.visibilityState === 'visible') {
+        try {
+          const hasSession = await checkSessionAction();
+          if (!hasSession) {
+            throw new Error('Session expired');
+          }
+        } catch (error) {
+          window.location.reload();
+        }
+      }
+    };
+
+    window.addEventListener('visibilitychange', handleVisibilityChange);
+
+    return () => {
+      window.removeEventListener('visibilitychange', handleVisibilityChange);
+    };
+  }, []);
+
+  return <>{children}</>;
+};
--- a/src/provider.tsx
+++ b/src/provider.tsx
@@ -3,7 +3,7 @@
 import * as React from 'react';
 import { checkSessionAction } from './actions.js';
 
-export const Provider = ({ children }: React.PropsWithChildren) => {
+export const AuthKitProvider = ({ children }: React.PropsWithChildren) => {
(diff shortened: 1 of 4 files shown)


src/provider.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+
+import * as React from 'react';
+import { checkSessionAction } from './actions.js';
+
+export const Provider = ({ children }: React.PropsWithChildren) => {
```
PaulAsjes, line 6: We should rename this to be more specific, like `AuthKitProvider` or `AuthProvider`.

src/provider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  React.useEffect(() => {
+    const handleVisibilityChange = async () => {
+      if (document.visibilityState === 'visible') {
+        try {
+          const hasSession = await checkSessionAction();
+          if (!hasSession) {
+            throw new Error('Session expired');
+          }
+        } catch (error) {
+          window.location.reload();
```
PaulAsjes, line 16: Should this be configurable? What if a developer wants to take an action other than reloading the page?

src/provider.tsx
```diff
@@ -0,0 +1,29 @@
...
+        }
+      }
+    };
+
+    window.addEventListener('visibilitychange', handleVisibilityChange);
```
PaulAsjes, line 21: We should leave a comment on why we're doing this and the edge case we're trying to solve.

PaulAsjes (changes requested):

PaulAsjes (approved):

PR #81, opened 2024-09-02
Version bump

PaulAsjes:


src/workos.ts
```diff
@@ -1,7 +1,7 @@
 import { WorkOS } from '@workos-inc/node';
 import { WORKOS_API_HOSTNAME, WORKOS_API_HTTPS, WORKOS_API_KEY, WORKOS_API_PORT } from './env-variables.js';
 
-export const VERSION = '0.9.0';
+export const VERSION = '0.10.0';
```
lucasmotta, line 4: In the future we should infer this from the package.json version.

lucasmotta (approved):

src/workos.ts
PaulAsjes: If we do that (require from `package.json`) it breaks the build step by trying to pull in package.json and the src file structure. Trust me, it's easier to update this here.

PR #82, opened 2024-09-03
Allow for dynamically set redirect URIs

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/45 https://github.com/workos/authkit-nextjs/issues/58 https://github.com/workos/authkit-nextjs/issues/24

This PR does a few things:

* Rename `getUser` to `withAuth` to be more in line with other SDKs and be more accurate in general
* Make `WORKOS_COOKIE_NAME` configurable in env variables
* Redirect URI can now be set dynamically, useful for things like Vercel preview deployments
* Removes the checks for environment variables as the benefit wasn't worth the added frustration


moroshko:
> Redirect URI can now be set dynamically, useful for things like Vercel preview deployments

@PaulAsjes Any plans to support dynamic logout redirect urls (the app homepage field in the Dashboard)? Without this ability, Vercel preview deployments logout flow can't work.

PaulAsjes:
@moroshko That's something we're considering, but don't have a timeline to share at the moment. I've passed it along to the team though!

package.json
```diff
@@ -21,7 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "@workos-inc/node": "7.21.0",
+    "@workos-inc/node": "7.26.0",
```
hadihallak, line 24: Should this be a Major release since it includes breaking changes?

src/env-variables.ts
```diff
@@ -1,37 +1,27 @@
-function getEnvVariable(name: string): string {
-  const envVariable = process.env[name];
-  if (!envVariable) {
-    throw new Error(`${name} environment variable is not set`);
-  }
-  return envVariable;
```
hadihallak, line 6: Out of curiosity, what's the reasoning behind this change?

src/session.ts
```diff
@@ -195,13 +208,15 @@ function getMiddlewareAuthPathRegex(pathGlob: string) {
...
 }
 
-async function getUser(options?: { ensureSignedIn: false }): Promise<UserInfo | NoUserInfo>;
-async function getUser(options: { ensureSignedIn: true }): Promise<UserInfo>;
-async function getUser({ ensureSignedIn = false } = {}) {
-  const session = await getSessionFromHeader('getUser');
+async function withAuth(options?: { ensureSignedIn: false }): Promise<UserInfo | NoUserInfo>;
+async function withAuth(options: { ensureSignedIn: true }): Promise<UserInfo>;
+async function withAuth({ ensureSignedIn = false } = {}) {
+  const session = await getSessionFromHeader('withAuth');
```
hadihallak, line 218: why do we pass the name of the function to the `getSessionFromHeader` function?

hadihallak (approved):
Haven't run the code locally but it looks good. Just left a few questions.

package.json
PaulAsjes: Since authkit-nextjs is pre v1.0 it's okay to have breaking changes in minor updates. See [point 4](https://semver.org/) on semver.

Also that line just bumps the version of workos-node, it's not the release version for this library.

src/session.ts
PaulAsjes: This is a bit of a legacy thing since we used to have more than one function that calls `getSessionFromHeader` and we wanted to be able to include the caller in the thrown error. I'll remove it since it's not used anymore.

src/env-variables.ts
PaulAsjes: We originally added this as a DX feature, but we've gotten reports that it's more annoying than helpful. See https://github.com/workos/authkit-nextjs/issues/45

Dvinge:
Those changes would solve a lot of problems we're having right now. Is there any potential ETA for getting this merged in?

ijxy:
Just a suggestion, but to work with Vercel deployment weirdness* it would be beneficial to use `NEXT_PUBLIC_` env variables where possible.

*: middleware cannot use variables from `.env` files unless they are `NEXT_PUBLIC_`.  For example, the way I use this is to add `NEXT_PUBLIC_WORKOS_REDIRECT_URI = "https://${NEXT_PUBLIC_VERCEL_URL}/some/path"` to the `.env.production` file (ie for previews and production builds).

PaulAsjes:
Sorry for the delay here, we're making a fairly significant change by [switching to esm from cjs](https://github.com/workos/authkit-nextjs/pull/87) and want to make sure that is stable before we merge. Since this PR also contains breaking changes we're waiting for that to land before merging this in.

PaulAsjes:
> Just a suggestion, but to work with Vercel deployment weirdness* it would be beneficial to use `NEXT_PUBLIC_` env variables where possible.
> 
> *: middleware cannot use variables from `.env` files unless they are `NEXT_PUBLIC_`. For example, the way I use this (with my hacked-up version of this library) is to add `NEXT_PUBLIC_WORKOS_REDIRECT_URI = "https://${NEXT_PUBLIC_VERCEL_URL}/some/path"` to the `.env.production` file (ie for previews and production builds).

@ijxy looking at the [Next.js docs](https://nextjs.org/docs/app/building-your-application/deploying#environment-variables), the `NEXT_PUBLIC_` prefix is only if you want to expose the env variable to the browser. Is there any reason why you'd want to expose the redirect URI?

ijxy:
> > Just a suggestion, but to work with Vercel deployment weirdness* it would be beneficial to use `NEXT_PUBLIC_` env variables where possible.
> > *: middleware cannot use variables from `.env` files unless they are `NEXT_PUBLIC_`. For example, the way I use this (with my hacked-up version of this library) is to add `NEXT_PUBLIC_WORKOS_REDIRECT_URI = "https://${NEXT_PUBLIC_VERCEL_URL}/some/path"` to the `.env.production` file (ie for previews and production builds).
> 
> @ijxy looking at the [Next.js docs](https://nextjs.org/docs/app/building-your-application/deploying#environment-variables), the `NEXT_PUBLIC_` prefix is only if you want to expose the env variable to the browser. Is there any reason why you'd want to expose the redirect URI?

This issue explains well: https://github.com/vercel/next.js/discussions/39705 

**tl;dr**: To expose custom env variables from any `.env` files to middleware (when hosting on Vercel) they need to be prefixed with `NEXT_PUBLIC_`.

ijxy:
@PaulAsjes as an example:

```
# .env
WORKOS_REDIRECT_URI_RUNTIME="https://${VERCEL_URL}/api/auth/callback"
NEXT_PUBLIC_WORKOS_REDIRECT_URI_RUNTIME="https://${VERCEL_URL}/api/auth/callback"
```

```
// middleware.ts

import { authkitMiddleware } from "@workos-inc/authkit-nextjs";
import { MiddlewareConfig, NextFetchEvent, NextRequest } from "next/server";

const auth = authkitMiddleware({
  middlewareAuth: { enabled: true, unauthenticatedPaths: [] },
});

export function middleware(request: NextRequest, event: NextFetchEvent) {
  console.log({
    WORKOS_REDIRECT_URI_RUNTIME: process.env.WORKOS_REDIRECT_URI_RUNTIME,
    NEXT_PUBLIC_WORKOS_REDIRECT_URI_RUNTIME: process.env.NEXT_PUBLIC_WORKOS_REDIRECT_URI_RUNTIME,
  });
  return auth(request, event);
}

export const config: MiddlewareConfig = {
  matcher: "/((?!api/|_next/|_vercel/|_assets/|[\\w-]+\\.\\w+).*)",
};

```

You'd expect both variables to be equivalent (since middleware runs in a secure environment) but what you actually see is:

```
{
  WORKOS_REDIRECT_URI_RUNTIME: undefined,
  NEXT_PUBLIC_WORKOS_REDIRECT_URI_RUNTIME: 'https://***-niq3m0chf-***.vercel.app/api/auth/callback'
}
```

The consequences of the above are:
- preview deployments don't work (you _need_ to be able to use VERCEL_URL with variable expansion, which means you need to be able to use a `.env` file, which means it needs to be `NEXT_PUBLIC_` due to the way Vercel works as demonstrated above)
- for production deployments, the domain has to be hardcoded into the env variable. Not the end of the world, but more error prone than relying on `VERCEL_URL`

These are very Vercel specific issues, but it is literally impossible to work around them using this package as it is due to the environment variable naming.

PaulAsjes:
@ijxy Thanks for that, that really clears it up. I've renamed `WORKOS_REDIRECT_URI` to `NEXT_PUBLIC_WORKOS_REDIRECT_URI` to unblock Vercel preview deployments.

PR #100, opened 2024-10-03
Fix some return types

PaulAsjes:
Upgrades the workos-node version, also fixes some inconsistencies with the return types of some of the helper methods.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -188,12 +188,12 @@ async function refreshSession({
 
   return {
     sessionId,
-    user: session.user,
+    user,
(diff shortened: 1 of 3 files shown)


src/session.ts
```diff
@@ -188,12 +188,12 @@ async function refreshSession({
 
   return {
     sessionId,
-    user: session.user,
```
PaulAsjes, line 191: Here we were accidentally returning the user, impersonator and accessToken from the _old_ session instead of the new one.

PR #102, opened 2024-10-04
Fix issue with redirectURI in middleware auth mode

PaulAsjes:
We set the redirectURI as a header, which isn't immediately available in middleware auth mode as it's the same request. Fix is to provide the redirect URI directly.

### DIFF
--- a/src/get-authorization-url.ts
+++ b/src/get-authorization-url.ts
@@ -4,9 +4,9 @@ import { GetAuthURLOptions } from './interfaces.js';
 import { headers } from 'next/headers';
 
 async function getAuthorizationUrl(options: GetAuthURLOptions = {}) {
-  const { returnPathname, screenHint, organizationId } = options;
+  const { returnPathname, screenHint, organizationId, redirectUri: redirectUriOptions } = options;
 
-  const redirectUri = headers().get('x-redirect-uri');
+  const redirectUri = redirectUriOptions ?? headers().get('x-redirect-uri');
 
   return workos.userManagement.getAuthorizationUrl({
     provider: 'authkit',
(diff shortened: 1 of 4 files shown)


src/get-authorization-url.ts
```diff
@@ -4,9 +4,9 @@ import { GetAuthURLOptions } from './interfaces.js';
 import { headers } from 'next/headers';
 
 async function getAuthorizationUrl(options: GetAuthURLOptions = {}) {
-  const { returnPathname, screenHint, organizationId } = options;
+  const { returnPathname, screenHint, organizationId, redirectUri: redirectUriOptions } = options;
```
cmatheson, line 7: nit: suggest singular `redirectUriOption` for the variable name. or... (maybe too fancy?):
```suggestion
  const { returnPathname, screenHint, organizationId, redirectUri = headers().get('x-redirect-uri') } = options;
```

src/session.ts
```diff
@@ -81,7 +81,9 @@ async function updateSession(
   if (middlewareAuth.enabled && matchedPaths.length === 0 && !session) {
     if (debug) console.log('Unauthenticated user on protected route, redirecting to AuthKit');
 
-    return NextResponse.redirect(await getAuthorizationUrl({ returnPathname: getReturnPathname(request.url) }));
+    return NextResponse.redirect(
+      await getAuthorizationUrl({ returnPathname: getReturnPathname(request.url), redirectUri: redirectUri ?? '' }),
```
cmatheson, line 85: this blank string is destined to fail? can we do something better?

cmatheson (approved):

PR #116, opened 2024-10-19
Fix `signOut` for custom cookie domains

KNWR:
The `signOut` method in auth.ts does not delete cookies set with a custom domain. This PR lets `signOut` delete custom domain cookies by checking if a custom domain has been set for the cookie. If so, it adds the custom domain to the keys that `cookies().delete()` (Nextjs' ResponseCookie's delete method) matches against, now finding the cookie and deleting it.

This lets authkit-nextjs work for applications that use a subdomain per customer (ex. `customer1.myenterpriseco.com`).


PaulAsjes (approved):
This is great, thanks!

Issue #104, opened 2024-10-05
NextJS 15 canary async headers/cookies compatibility


f1shy-dev:
``` In route / a cookie property was accessed directly with `cookies().get('wos-session')`. `cookies()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
 In route /ai-sdk-chat/[id] a header property was accessed directly with `headers().get('x-workos-middleware')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
 In route /ai-sdk-chat/[id] a header property was accessed directly with `headers().get('x-workos-session')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis```
 
Latest NextJS canary beta is showing these errors, using AuthKit and getUser()

Likely a simple change that needs to be made inside authkit-nextjs?

PaulAsjes:
Thanks for the report! We'll look into this. The fix should be easy enough, just want to make sure that awaiting that getter is backwards compatible with older versions of Next.

nirtamir2:
Now Next.js 15 is stable - so I think it's safe to change this and await the cookie:
https://github.com/workos/authkit-nextjs/blob/7fa52f555fdad62a2e76579d6739e067acbfd059/src/session.ts#L105


Made a small PR for you to check 
https://github.com/workos/authkit-nextjs/pull/118

PaulAsjes:
Thanks, we're working on this right now and hope to get a fix out today.

nirtamir2:
Thanks for the quick response! ❤️
So no need to patch-package 🎉

PaulAsjes:
This has been released as [v0.13.0](https://github.com/workos/authkit-nextjs/releases/tag/v0.13.0)!

PR #119, opened 2024-10-22
Next 15 compatibility

PaulAsjes:
Updates dependencies and adds Next 15 compatibility.

Fixes https://github.com/workos/authkit-nextjs/issues/104

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -15,7 +15,8 @@ async function getSignUpUrl() {
 
 async function signOut() {
   const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
-  cookies().delete(cookieName);
+  const nextCookies = await cookies();
+  nextCookies.delete(cookieName);
   await terminateSession();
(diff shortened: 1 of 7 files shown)


src/auth.ts
```diff
@@ -18,7 +18,9 @@ async function signOut() {
     name: WORKOS_COOKIE_NAME || 'wos-session',
   };
   if (WORKOS_COOKIE_DOMAIN) cookie.domain = WORKOS_COOKIE_DOMAIN;
-  cookies().delete(cookie);
+
+  const nextCookies = await cookies();
+  nextCookies.delete(cookie);
```
mattgd, line 23: Confirming this also works for Next 14?

mattgd (approved):

src/auth.ts
PaulAsjes: Tested that it works on the next-authkit-example app, which is still on Next 14.

Issue #121, opened 2024-10-25
Configure redirect URI screen hints


PaulAsjes:
Should allow the "sign up" screen hint to be configurable on a per-path basis. This way users who are redirected to AuthKit see the appropriate screen hint.

Example:
```ts
export default authkitMiddleware({ 
  signUpPaths: ['/organization/join', '/foo', '/bar']
});
```

PR #123, opened 2024-10-28
Add signUpPaths configuration

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/121

Adds `signUpPaths` to middleware configuration, which allows you to specify which paths should show the "sign up" screen hint over the default "sign in".

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -363,7 +363,8 @@ function getScreenHint(signUpPaths: string[] | string | undefined, pathname: str
   if (!signUpPaths) return 'sign-in';
 
   if (!Array.isArray(signUpPaths)) {
-    return signUpPaths.includes(pathname) ? 'sign-up' : 'sign-in';
+    const pathRegex = getMiddlewareAuthPathRegex(signUpPaths);
+    return pathRegex.exec(pathname) ? 'sign-up' : 'sign-in';
   }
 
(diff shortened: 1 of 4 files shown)


src/session.ts
```diff
@@ -344,4 +359,19 @@ function getReturnPathname(url: string): string {
...
+  if (!signUpPaths) return 'sign-in';
+
+  if (!Array.isArray(signUpPaths)) {
+    return signUpPaths.includes(pathname) ? 'sign-up' : 'sign-in';
+  }
+
+  const screenHintPaths: string[] = signUpPaths.filter((pathGlob) => {
+    const pathRegex = getMiddlewareAuthPathRegex(pathGlob);
+    return pathRegex.exec(pathname);
+  });
```
mattgd, line 372: Is there a reason we pass the paths through `getMiddlewareAuthPathRegex` if `signUpPaths` is an array, but not if it's a single value? Could a single value also be a path glob?

mattgd (commented):
Just have one question regarding `getScreenHint`.

src/session.ts
PaulAsjes: That's an excellent point!

mattgd (approved):

Issue #37, opened 2024-05-02
Post login redirect taking me to 0.0.0.0:3000


elliotcw:
After signing in with WorkOS I am redirected to 0.0.0.0:3000.

Our deployed Next.js application is built and runs in a docker container. We run the container using Google Cloud Run. It looks like the authkit-nextjs middleware is using the `request.nextUrl` when handling the auth callback.

My suspicion is `request.nextUrl` is resolving to an internal address, 0.0.0.0:3000.

Our docker container has the following set:
```
EXPOSE 3000

ENV PORT 3000
# set hostname to localhost
ENV HOSTNAME "0.0.0.0"

# server.js is created by next build from the standalone output
# https://nextjs.org/docs/pages/api-reference/next-config-js/output
CMD ["node", "apps/example-app/server.js"]
```

Is there a way to override this behaviour. Any pointers to how I can deploy my next.js application in docker and have workos redirects work correctly?

elliotcw:
I was able to work around it using the Host header, based on this solution https://github.com/vercel/next.js/issues/37536#issuecomment-1157000990

Adding it to the middleware though didnt work out. I needed to add it to the auth/callback route handler.

```
// app/auth/callback/route.tsx
import { handleAuth } from '@workos-inc/authkit-nextjs';

// Redirect the user to `/` after successful sign in
// The redirect can be customized: `handleAuth({ returnPathname: '/foo' })`
const handler = handleAuth();

export const GET = async (request) => {
  const { nextUrl, headers } = request;
  let { url } = request;
  nextUrl.host = headers.get('Host') ?? nextUrl.host;

  // clear the port if not localhost
  if (!nextUrl.host.includes('localhost')) {
    nextUrl.port = '';
  }

  url = nextUrl.href;

  return handler(request);
};
```

raphaelcunha:
I have the same situation.

My suggestion is create a new prop to use a baseUrl when necessary for this cases.

`export const GET = handleAuth({
  returnPathname: "/dashboard",
  baseUrl: "https://my-domain.com"
});`

mousetree:
Having same issue

PaulAsjes:
Hi all and thanks for your patience. We have a [PR up](https://github.com/workos/authkit-nextjs/pull/130) that should fix this (thanks @raphaelcunha for the suggestion!) that I'm hoping to release later this week.

PaulAsjes:
I've released the above as v0.14.0. Let me know if you're still having issues!

[...]

PR #130, opened 2024-11-06
Adds configurable baseURL in the auth redirect route

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/37

Adds a new `baseURL` configuration option to `handleAuth`. This is to circumvent scenarios where `request.nextUrl` points to an internal URL rather than the intended one.


mattgd (approved):
Looks sane to me 👍

PR #132, opened 2024-11-09
 README.md - Add correct types to authkitMiddleware function.

raphaelcunha:
### Add correct types to `authkitMiddleware` function

The `middleware` function requires two parameters, but this wasn't clearly documented. Updated the function to ensure the second parameter, `event`, is included.

**Before (incorrect)**  
❌  
CleanShot 2024-11-09 at 12 36 48@2x

**After (correct)**  
✅  
```typescript
export default async function middleware(request: NextRequest, event: NextFetchEvent) {
  const response = await authkitMiddleware()(request, event); // added event param.
  return response;
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,11 +212,11 @@ Sometimes it's useful to check the user session if you want to compose custom mi
 
 ```ts
 import { authkitMiddleware, getSession } from '@workos-inc/authkit-nextjs';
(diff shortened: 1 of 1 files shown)


PaulAsjes (approved):
Thanks!

PR #143, opened 2024-11-21
Redirect if refresh failed

PaulAsjes:


marji-workos (approved):

src/session.ts
```diff
@@ -99,7 +99,7 @@ async function updateSession(
     return NextResponse?.redirect
       ? NextResponse.redirect(redirectTo)
       : new Response(null, {
-          status: 302,
+          status: 307,
```
cmatheson, line 102: should this be a 307? if the original request were POSTing to some endpoint, i think a 307 would have you POST (the original request body) to the authorization url (but we want to GET).

src/session.ts
```diff
@@ -160,12 +160,22 @@ async function updateSession(
...
-    return response;
+
+    nextCookies.delete(cookieName);
   }
+
+  // If we get here, the session is invalid and the user needs to sign in again.
+  // We redirect to the current URL which will trigger the middleware again.
+  // This is outside of the above block because you cannot redirect in Next.js
+  // from inside a try/catch block.
+  return NextResponse?.redirect
```
cmatheson, line 171: why not redirect to getAuthorizationUrl directly?

cmatheson (approved):

Issue #146, opened 2024-12-08
Component-level authentication


ibagov:
Is there a way to authenticate client-side components at the component level, as opposed to the route level?

I have a modal form in my Next app which can be opened from almost any route, however the form itself is not associated with a specific route. Can I ensure that only authorized users are able to access it? Obviously, using the `AuthKitMiddleware` is not applicable in this scenario.

PaulAsjes:
If you're using server components, then calling `withAuth` within your component will give you either the user object or `null` if the user isn't authenticated. From there you should be able to either render a logged-out view or force the user to sign in (which you can also do by passing in `ensureSignedIn: true` to the `withAuth` call.

We're currently thinking about adding a way to do the above in a client component. Let me know if that helps!

ibagov:
Understood, thanks for the reply @PaulAsjes . So if I understand correctly, it's currently not possible to solve my problem using `authkit-nextjs`.

Do you think it make sense to try using `workos-node` for this use case?

PaulAsjes:
You can if you're using a server component with `withAuth`. If you're using a client component then you can't just yet, but I'm currently working on a client side solution to getting auth. Hoping to have it released sometime next week.

ibagov:
Got it, thanks and looking forward to the solution! :)

sergesteban:
> You can if you're using a server component with `withAuth`. If you're using a client component then you can't just yet, but I'm currently working on a client side solution to getting auth. Hoping to have it released sometime next week.


I also faced the same issue while using auth on component-level, and I ended creating an extra context so I could make use of it. That felt redundant, tho I am glad to noticed the changes on #147  @PaulAsjes  thanks 
 
looking forward to it
cheers

ibagov:
Sorry to reopen the issue again, but the update does not seem to have fixed my issue/feature request.

If I use the `useAuth` hook inside of a client **component** like so:

 `const { user, loading } = useAuth({ ensureSignedIn: true });`

the component will still get rendered without an automatic redirect to the sign-in page, which I think would be the expected behavior.

Issue #80, opened 2024-09-01
impersonation code breaks the build


dror-trail:
using Next.js (14.2.5) + "@workos-inc/authkit-nextjs": "^0.9.0"
i dont use the impersonation component but it still breaks

          Error: ./node_modules/@workos-inc/authkit-nextjs/dist/cjs/impersonation.js:54:21
          Ecmascript file had an error
            52 |             } },
            53 |             React.createElement("form", { action: async () => {
          > 54 |                     'use server';
               |                     ^^^^^^^^^^^^
            55 |                     await (0, auth_js_1.signOut)();
            56 |                 }, style: {
            57 |                     display: 'flex',
          
          It is not allowed to define inline "use server" annotated Server Actions in Client Components.
          To use Server Actions in a Client Component, you can either export them from a separate file with "use server" at the top, or pass them down through props from a Server Component.

PaulAsjes:
I'm unable to reproduce with both Next.js 14.2.5 and authkit-next 0.9.0. Could you provide more details or an example repo that shows the issue?

MHendrieAIOT:
I'm having this same problem. I need to use signOut() in an onClick event handler (outside of a form) and I haven't been able to get it to work.

Some minimal code that throws this error for me:
```ts
"use client"
import { signOut } from "@workos-inc/authkit-nextjs";
export default async function Example() {
  return (
    <button onClick={async () => await signOut()}>Sign out</button>
  );
}
```
The error still occurs if I extract the signOut call to a server action in a separate file and import it in my client component.

PaulAsjes:
Hi all, we've found that the issue only appears in client components. We're working on a fix, but in the meanwhile for client components you should be able to treat `signOut` and `signIn` as a server function:

```jsx
"use client"
import { signOut } from "@workos-inc/authkit-nextjs";
export default async function Example() {
  return (
    <button onClick={signOut}>Sign out</button>
  );
}
```

`Impersonation` and `AuthKitProvider` won't currently work in client components, but we're working on a fix.

igorsmotto:
Hey @PaulAsjes!

Is there any ETA or workaround for this issue?

kmankan:
is there a solution for this yet? this means a bunch of client side function like withAuth don't work either... that's pretty important.

monolithed:
I can't check the behavior in the latest version because it's broken. In version 0.16.2, when calling <Impersonation/> in a server or client component, the following error occurs:

[...]

PR #147, opened 2024-12-10
Add client side auth

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/80
Hopefully fixes https://github.com/workos/authkit-nextjs/issues/146

Introduces new auth hooks via the `AuthKitProvider` component. Also separates server and client components so no crypto modules are accidentally included, causing errors in client components that attempt to use this library.


chamini2:
Hi is there any timeline for this? I am blocked to integrate into our existing app without major refactors.

monolithed:
It’d be awesome to also show an implementation of Magic Auth using `workos.userManagement.authenticateWithMagicAuth` 🙂

README.md
```diff
@@ -200,16 +200,84 @@ export default async function HomePage() {
...
+
+For client components, use the `useAuth` hook to get the current user session.
+
+```jsx
+// Note the updated import path
+import { useAuth } from '@workos-inc/authkit-nextjs/components';
+
+export default function MyComponent() {
+  // Retrieves the user from the session or returns `null` if no user is signed in
+  const { user, loading } = useAuth();
```
jessedijkstra, line 213: @PaulAsjes this function no longer takes `{ensureSignedIn: true}` for it's arguments.

README.md
```diff
@@ -200,16 +200,84 @@ export default async function HomePage() {
...
 ### Requiring auth
 
 For pages where a signed-in user is mandatory, you can use the `ensureSignedIn` option:
 
 ```jsx
+// Server component
+
+// Client component
+const { user, loading } = useAuth({ ensureSignedIn: true });
```
jessedijkstra, line 232: @PaulAsjes this doesn't work after the upgrade to version 1.x

README.md
PaulAsjes: This and the comment below were fixed in v1.0.2.

jessedijkstra: Thanks!

Issue #112, opened 2024-10-16
Missing cookie password throws a confusing error


PaulAsjes:
The callback route logic throws an unhelpful 
```json
{ error: Empty password` }
```
error if the cookie password is missing from the env variables.

sergesteban:
I would also suggest to update this error message
```
{ error: 'Password string too short (min 32 characters required)' }
```

I didn’t actually notice at first, but I was missing characters in WORKOS_COOKIE_PASSWORD. I think the problem here is DX imo, since everything was flowing smoothly, and I didn’t notice the missing characters. 
I regenerated the password, and it started working again after that. I would’ve appreciated something like a hint on `Cookie password string...`.

thanks!

PR #153, opened 2024-12-16
Better cookie error handling

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/112

Throws if the cookie password is missing or if it's too short.

Also adds a test for the `onSuccess` handler, which was merged prior to when we had tests for this repo.

### DIFF
--- a/__tests__/authkit-callback-route.spec.ts
+++ b/__tests__/authkit-callback-route.spec.ts
@@ -16,6 +16,28 @@ jest.mock('../src/workos', () => ({
 }));
 
 describe('authkit-callback-route', () => {
+  const mockAuthResponse = {
+    accessToken: 'access123',
+    refreshToken: 'refresh123',
+    user: {
+      id: 'user_123',
+      email: 'test@example.com',
+      emailVerified: true,
+      profilePictureUrl: 'https://example.com/photo.jpg',
+      firstName: 'Test',
+      lastName: 'User',
+      object: 'user' as const,
+      createdAt: '2024-01-01T00:00:00Z',
+      updatedAt: '2024-01-01T00:00:00Z',
+    },
+    oauthTokens: {
+      accessToken: 'access123',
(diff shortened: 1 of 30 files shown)


__tests__/authkit-callback-route.spec.ts
```diff
@@ -232,5 +232,30 @@ describe('authkit-callback-route', () => {
...
+        user: { id: 'user_123' },
+        oauthTokens: {
+          access_token: 'access123',
+          refresh_token: 'refresh123',
+          expires_at: 1719811200,
+          scopes: ['foo', 'bar'],
+        },
+      };
+
+      (workos.userManagement.authenticateWithCode as jest.Mock).mockResolvedValue(mockAuthResponse);
```
mthadley, line 249: [`jest.mocked`](https://jestjs.io/docs/mock-function-api#jestmockedsource-options) was intended to help avoid this type of casting and also retain the type information from the mocked method.

```suggestion
      jest.mocked(workos.userManagement.authenticateWithCode).mockResolvedValue(mockAuthResponse);
```

__tests__/session.spec.ts
```diff
@@ -159,6 +159,52 @@ describe('session.ts', () => {
...
+            unauthenticatedPaths: [],
+          },
+          '',
+          [],
+        );
+      }).rejects.toThrow(
+        'You must provide a valid cookie password that is at least 32 characters in the environment variables.',
+      );
+
+      jest.replaceProperty(envVariables, 'WORKOS_COOKIE_PASSWORD', originalWorkosCookiePassword);
```
mthadley, line 182: If you set `restoreMocks` to true in the Jest config then properties like this will automatically be restored between tests without a need to keep track of original values like this. I think it's a good default behavior to go with.

mthadley (approved):
LGTM! Just a few test-related nits and suggestions.

PR #154, opened 2024-12-17
Move `oauthTokens` out of `Session` and make available via a new `onSuccess` option

mthadley:
This follows the pattern that was recently introduced in workos/authkit-remix#33:

> Instead of storing these tokens in the session, this branch proposes making them only available as part of the initial > callback. Developers who need access to the underlying OAuth provider's tokens and API will be responsible for persisting them in their own data store for later usage.

The equivalent usage with `workos/authkit-next` looks like this:

```ts
import { handleAuth } from '@workos-inc/authkit-nextjs';

export const GET = handleAuth({
  onSuccess: async ({ oauthTokens }) => {
    await saveTokens(oauthTokens);
  },
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,23 @@ You can also control the pathname the user will be sent to after signing-in by p
 export const GET = handleAuth({ returnPathname: '/dashboard' });
 ```
 
-`handleAuth` can be used with several options.
+If you're application needs to persist data upon a successful authentication like the `oauthTokens` from an upstream provider, you can pass the `onSuccess` option:
+
+```ts
+export const GET = handleAUth({
+  onSuccess: async ({ oauthTokens }) => {
+    await saveTokens(oauthTokens);
+  },
+});
+```
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -80,12 +80,23 @@ You can also control the pathname the user will be sent to after signing-in by p
 export const GET = handleAuth({ returnPathname: '/dashboard' });
 ```
 
-`handleAuth` can be used with several options.
+If you're application needs to persist data upon a successful authentication like the `oauthTokens` from an upstream provider, you can pass the `onSuccess` option:
```
PaulAsjes, line 83:
```suggestion
If your application needs to persist data upon a successful authentication, like the `oauthTokens` from an upstream provider, you can pass in a `onSuccess` function that will get called after the user has successfully authenticated:
```

PaulAsjes (changes requested):
Looks good, just some wordsmithing.

PaulAsjes (approved):

Issue #170, opened 2025-01-13
authkit middleware not working on Next 14.2


chamini2:
![image](https://github.com/user-attachments/assets/50c85326-8801-403d-a26d-ef4c25c300ea)

my middleware.ts is similar to

```ts
import { authkit } from "@workos-inc/authkit-nextjs";
import { NextFetchEvent, NextRequest, NextResponse } from "next/server";
import {
  CSRF_COOKIE_NAME
} from "./lib/csrf";

async function csrfMiddleware(request: NextRequest, response: NextResponse) {
  const signature = "...";
  response.cookies.set(CSRF_COOKIE_NAME, signature, {
    sameSite: "strict",
    httpOnly: true,
  });
  response.headers.set("X-Frame-Options", "SAMEORIGIN");
}

export async function middleware(request: NextRequest, event: NextFetchEvent) {
  // This method will automatically handle setting the cookie and refreshing the session
  const auth = await authkit(request, {
    debug: false,
  });

  // Headers need to be included in every non-redirect response to ensure that `withAuth` works as expected
  const response = NextResponse.next({
    headers: auth.headers,
  });

  await csrfMiddleware(request, response);

  return response;
}

export const config = {
  matcher: [
    "/",
    "/demos",
    "/demos/:path*",
  ],
};

```

fierysolid:
Same issue

chamini2:
seeing this with debug

```
Session invalid. Refreshing access token that ends in qrng4KCikQ
Failed to refresh. Deleting cookie. [Error: Cookies can only be modified in a Server Action or Route Handler. Read more: https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-value-options]
```

chamini2:
only happens when refreshin

chamini2:
this is happening for the non composable solution too for me

```
export default authkitMiddleware({ debug: true });
```

PaulAsjes:
Hi all, I was able to recreate this. Will put up a fix later today.

PR #171, opened 2025-01-14
Fix cookie bugs

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/170

v1 shipped with bugs surrounding the refresh logic.

tl;dr is that we were using Next's `cookies()` helper method, which doesn't work in edge middleware. Instead we now set cookies via the headers. 

Want to release this as a fix as the general API is unchanged.

### DIFF
--- a/src/authkit-callback-route.ts
+++ b/src/authkit-callback-route.ts
@@ -5,7 +5,7 @@ import { WORKOS_CLIENT_ID, WORKOS_COOKIE_NAME } from './env-variables.js';
 import { encryptSession } from './session.js';
 import { errorResponseWithFallback, redirectWithFallback } from './utils.js';
 import { getCookieOptions } from './cookie.js';
-import { HandleAuthOptions } from './interfaces.js';
+import { CookieOptions, HandleAuthOptions } from './interfaces.js';
 
 export function handleAuth(options: HandleAuthOptions = {}) {
   const { returnPathname: returnPathnameOption = '/', baseURL, onSuccess } = options;
@@ -73,7 +73,7 @@ export function handleAuth(options: HandleAuthOptions = {}) {
         const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
         const nextCookies = await cookies();
 
-        nextCookies.set(cookieName, session, getCookieOptions(request.url));
+        nextCookies.set(cookieName, session, getCookieOptions(request.url) as CookieOptions);
 
         return response;
       } catch (error) {
--- a/src/authkit-callback-route.ts
+++ b/src/authkit-callback-route.ts
@@ -5,7 +5,7 @@ import { WORKOS_CLIENT_ID, WORKOS_COOKIE_NAME } from './env-variables.js';
 import { encryptSession } from './session.js';
 import { errorResponseWithFallback, redirectWithFallback } from './utils.js';
 import { getCookieOptions } from './cookie.js';
-import { CookieOptions, HandleAuthOptions } from './interfaces.js';
+import { HandleAuthOptions } from './interfaces.js';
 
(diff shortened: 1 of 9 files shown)


src/authkit-callback-route.ts
```diff
@@ -73,7 +73,7 @@ export function handleAuth(options: HandleAuthOptions = {}) {
         const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
         const nextCookies = await cookies();
 
-        nextCookies.set(cookieName, session, getCookieOptions(request.url));
+        nextCookies.set(cookieName, session, getCookieOptions(request.url) as CookieOptions);
```
nicknisi, line 76: **nit(non-blocking):** I think there's a way to improve the DX on `getCookieOptions` where the type could be automatically narrowed, but definitely not a blocker here.

src/cookie.ts
```diff
@@ -1,18 +1,26 @@
 import { WORKOS_REDIRECT_URI, WORKOS_COOKIE_MAX_AGE, WORKOS_COOKIE_DOMAIN } from './env-variables.js';
 import { CookieOptions } from './interfaces.js';
 
-export function getCookieOptions(redirectUri?: string | null): CookieOptions {
+export function getCookieOptions(
+  redirectUri?: string | null,
+  asString: boolean = false,
+  expired: boolean = false,
+): CookieOptions | string {
```
nicknisi, line 8: **suggestion(non-blocking):** Possible to overload the function signature to ensure the type is narrowed here.

```suggestion
export function getCookieOptions(): CookieOptions;
export function getCookieOptions(redirectUri?: string | null): CookieOptions;
export function getCookieOptions(redirectUri: string | null | undefined, asString: true, expired?: boolean): string;
export function getCookieOptions(
  redirectUri: string | null | undefined,
  asString: false,
  expired?: boolean,
): CookieOptions;
export function getCookieOptions(
  redirectUri?: string | null,
  asString?: boolean,
  expired?: boolean,
): CookieOptions | string;
export function getCookieOptions(
  redirectUri?: string | null,
  asString: boolean = false,
  expired: boolean = false,
): CookieOptions | string {
```

nicknisi (approved):
Changes look good to me. I added a small suggestion but don't consider it a blocker.

src/session.ts
```diff
@@ -143,69 +144,98 @@ async function updateSession(
...
+      headers: newRequestHeaders,
+    };
+  } catch (e) {
+    if (options.debug) {
+      console.log('Failed to refresh. Deleting cookie.', e);
+    }
+
+    // When we need to delete a cookie, return it as a header as you can't delete cookies from edge middleware
+    const deleteCookie = `${cookieName}=; Expires=${new Date(0).toUTCString()}; ${getCookieOptions(request.url, true, true)}`;
+    newRequestHeaders.append('Set-Cookie', deleteCookie);
```
bdbergeron, line 229: I think this attempt to delete a cookie doesn't work when developing locally - when the session is invalid and a refresh attempt fails, I'm still seeing the cookie in my browser's dev tools with an encoded value and expiration set over a year in the future.

The error is being logged as follows:
```
Session invalid. Refreshing access token that ends in hGt09AzLTA
Failed to refresh. Deleting cookie. OauthException: Error: invalid_grant
Error Description: Refresh token already exchanged.
    at Generator.throw (<anonymous>) {
  status: 400,
  requestID: '',
  error: 'invalid_grant',
  errorDescription: 'Refresh token already exchanged.',
  rawData: {
    error: 'invalid_grant',
    error_description: 'Refresh token already exchanged.'
  }
}
```

axanthus:
@bdbergeron I'm experiencing this exact issue. What code did you write to delete the cookie and solve the issue? For some reason my cookies are empty when I log them.

Issue #175, opened 2025-01-15
useAuth always returns null user


papaymaguire:
**Describe the bug**
attempting to get the user on a client component using useAuth() always returns null for the user.

**To Reproduce**
1. Basic Layout with AuthKitProvider as in the Getting Started Tutorial.
2. Setup callback function as in Getting Started Tutorial.
3. Middleware function:
```ts
import { authkitMiddleware } from '@workos-inc/authkit-nextjs';

export default authkitMiddleware({middlewareAuth: {enabled: true, unauthenticatedPaths: ['/api/auth/callback']}});
```
4. 
```ts
'use client'

import { useAuth } from "@workos-inc/authkit-nextjs/components"

export default function MembersPage() {
    const result = useAuth()

    if (result.loading) {
        return (
            <h1>Loading...</h1>
        )
    } else {
        return (
            <h1>{JSON.stringify(result)}</h1>
        )
    }
}
```

**Expected behavior**
After initial loading time, the members page should show the user object on the screen. Instead I get the following:
![image](https://github.com/user-attachments/assets/5d7f1034-8afe-4c60-aff1-52d6b7d50ae9)

**Context**
I am indeed logged in, the wos-session cookie is there and the route is protected with the middleware so I wouldn't be able to see anything if I wasn't. I'm not a NextJs expert but I also don't see anything in the network tab that would indicate a server action being called (if they would show up in the network calls).

PaulAsjes:
Hi there, I see the problem and am working on a fix. Should be out later today.

papaymaguire:
Awesome thank you, will this be released today?

PR #177, opened 2025-01-15
Fix server actions

PaulAsjes:
Fixes https://github.com/workos/authkit-nextjs/issues/175

We previously were copying headers from the request via `new Headers(request.headers)` which it turns out mangles the required headers for server actions.

This cleans up some types and makes sure we don't mess with the headers.

See this [issue](https://github.com/vercel/next.js/issues/50659#issuecomment-2333990159) for more details.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -21,8 +21,8 @@ export const getOrganizationAction = async (organizationId: string) => {
   return await workos.organizations.getOrganization(organizationId);
 };
 
-export const getAuthAction = async (ensureSignedIn?: boolean) => {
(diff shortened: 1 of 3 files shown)


nicknisi (approved):

moroshko:
@PaulAsjes I'm curious why the return type of `withAuth` when `ensureSignedIn` is `true` was modified to include `null`?

**v1.0.1**
```
// user is User
```

**v1.0.2**
```
// user is User | null
```

nicknisi:
I think this part should be able to be reversed since Next will throw an error to redirect if the user isn't signed in, when `ensureSignedIn` is true. Addressed in #185.

PR #183, opened 2025-01-17
Add client component signout method

### LINKED ISSUE #180: Add `signOut` to the client functions
Quite a bit of my app is done in client components behind auth. It would be great if there was a provided function to destroy the session cookie from the client since server components can't be placed inside of client components.

nicknisi:
- **Add signOut for client component**
- **fix act warnings in new signout tests**

Fixes #180

### DIFF
--- a/__tests__/impersonation.spec.tsx
+++ b/__tests__/impersonation.spec.tsx
@@ -1,9 +1,10 @@
-import { render, act } from '@testing-library/react';
+import { render, act, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { Impersonation } from '../src/components/impersonation.js';
 import { useAuth } from '../src/components/authkit-provider.js';
 import { getOrganizationAction } from '../src/actions.js';
 import * as React from 'react';
+import { handleSignOutAction } from '../src/actions.js';
 
 // Mock the useAuth hook
 jest.mock('../src/components/authkit-provider', () => ({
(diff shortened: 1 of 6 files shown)


__tests__/impersonation.spec.tsx
```diff
@@ -116,4 +117,18 @@ describe('Impersonation', () => {
...
+  it('should should sign out when the Stop button is called', async () => {
+    (useAuth as jest.Mock).mockReturnValue({
+      impersonator: { email: 'admin@example.com' },
+      user: { id: '123', email: 'user@example.com' },
+      organizationId: null,
+      loading: false,
+    });
+
+    render(<Impersonation />);
+    const stopButton = await screen.findByText('Stop');
```
cmatheson, line 130: nit: i was confused why we were clicking 'Stop' instead of a "log out"/signout button. (is there a non-impersonation spec that we could [additionally?] add this to?

cmatheson (approved):

__tests__/impersonation.spec.tsx
nicknisi: Is this here what you meean?

https://github.com/workos/authkit-nextjs/pull/183/files#diff-a8c3e17afcc9d081a13c30f8987cc7d7c70b8fb58c784e09238edb9344224749R398

cmatheson: oops yeah that's what i'm talking about

PR #185, opened 2025-01-18
Narrow withAuth type to User when ensureSignedIn is true

nicknisi:
When `ensureEignedIn` is true, the user will either be returned or next will throw a `NEXT_REDIRECT` to redirect to log in. This is described in #177.

```ts
const { user } = await withAuth(); // user is User or null

const { user } = await withAuth({ ensureSignedIn: false }); // user is User or null

const { user } = await withAuth({ ensureSignedIn: trie }); // user is User
```

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -351,12 +351,13 @@ async function redirectToSignIn() {
   redirect(await getAuthorizationUrl({ returnPathname, screenHint }));
 }
 
-async function withAuth(options?: { ensureSignedIn?: boolean }): Promise<UserInfo | NoUserInfo>;
-async function withAuth({ ensureSignedIn = false }: { ensureSignedIn?: boolean } = {}): Promise<UserInfo | NoUserInfo> {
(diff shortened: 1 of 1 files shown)


moroshko:
@nicknisi Can we do the same with `useAuth()`?

```
const { user, loading } = useAuth({ ensureSignedIn: true });

if (loading) {
  // ...
} else {
  console.log(user); // Currently, the type here is User | null. Can it be User?
}
```

nicknisi:
@moroshko Hmm, this seems reasonable and I think can be accomplished with a simple function overload. When `ensureSignedIn: true` and `loading: false`, I think we can assume that `user` won't be null. I gave it a shot in #196.

moroshko:
@nicknisi Is it possible to narrow the type here as well?

```ts
import { refreshSession } from "@workos-inc/authkit-nextjs";

const { accessToken } = await refreshSession({ ensureSignedIn: true });
// accessToken should be string, not string | undefined
```

nicknisi:
@moroshko Yes! I submitted #215 and will get it into the next release. Thanks!

PR #198, opened 2025-02-10
Fix interfaces.d.ts not being found

### LINKED ISSUE #181: Interfaces cannot be found in 1.1.0
**Describe the bug**
Types are not resolvable in 1.1.0

**To Reproduce**
install package, attempt to import a type

**Additional context**
Types are located in ./dist/esm/types/interfaces.ts but not exported in package.json

spookyuser:
This fixes #181 for me

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,10 +20,6 @@
     ".": {
       "types": "./dist/esm/types/index.d.ts",
       "import": "./dist/esm/index.js"
-    },
-    "./interfaces": {
-     "types": "./dist/esm/types/interfaces.d.ts",
-     "import": "./dist/esm/interfaces.js"
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


nicknisi:
@spookyuser Thanks for the contribution! I'm wondering if it would also solve your needs if the types were exported through the code?

```ts
// inside src/index.ts

export * from './interfaces.js';
```

This way, you'd be able to access the types straight from the main package import:

```ts
import { type UserInfo } from "@workos-inc/authkit-nextjs";
```

spookyuser:
No problem! Sure I think that would work let me just double check

spookyuser:
It worked great thank you, I updated the pr with that export instead of changing the package.json

nicknisi (approved):

nicknisi:
@spookyuser Thanks! This is included in v1.4.0.

PR #206, opened 2025-02-21
remove accessToken from being sent to client components

nicknisi:
**Breaking change**

This removes `accessToken` from being accessible in client components. The following is no longer allowed.

```typescript
const { accessToken } = useAuth(); // no longer accessible
```

This pull request focuses on removing the `accessToken` from the client-side code to enhance security and simplify the auth object. The most important changes include sanitizing the auth object, updating actions to use the sanitized auth object, and modifying components to remove references to `accessToken`.

### Security Improvements:
* Added a `sanitize` function to remove the `accessToken` from the auth object before it is used on the client side. (`src/actions.ts`)
* Updated `getAuthAction` and `refreshAuthAction` to use the `sanitize` function. (`src/actions.ts`) [[1]](diffhunk://#diff-1b65654b4e6a4ad772c76520a607abec72eb17b56caaace589a3c5c51a21fdcbL25-R36) [[2]](diffhunk://#diff-1b65654b4e6a4ad772c76520a607abec72eb17b56caaace589a3c5c51a21fdcbL35-R46)

### Component Updates:
* Removed `accessToken` from the `AuthContextType` type definition. (`src/components/authkit-provider.tsx`)
* Removed `accessToken` state and related code from the `AuthKitProvider` component. (`src/components/authkit-provider.tsx`) [[1]](diffhunk://#diff-5ff17fea1fc95fa55adfd7e345657a5df19aa7ebfd3ba619f66fda027b523041L41) [[2]](diffhunk://#diff-5ff17fea1fc95fa55adfd7e345657a5df19aa7ebfd3ba619f66fda027b523041L54) [[3]](diffhunk://#diff-5ff17fea1fc95fa55adfd7e345657a5df19aa7ebfd3ba619f66fda027b523041L63) [[4]](diffhunk://#diff-5ff17fea1fc95fa55adfd7e345657a5df19aa7ebfd3ba619f66fda027b523041L84) [[5]](diffhunk://#diff-5ff17fea1fc95fa55adfd7e345657a5df19aa7ebfd3ba619f66fda027b523041L157)

### Test Adjustments:
* Removed `accessToken` from test mock data in `useAuth` test suite. (`__tests__/authkit-provider.spec.tsx`)

### DIFF
--- a/__tests__/authkit-provider.spec.tsx
+++ b/__tests__/authkit-provider.spec.tsx
@@ -219,7 +219,6 @@ describe('useAuth', () => {
       permissions: ['read', 'write'],
       entitlements: ['feature1'],
       impersonator: { email: 'admin@example.com' },
-      oauthTokens: { access_token: 'token123' },
     });
 
     const TestComponent = () => {
--- a/__tests__/authkit-provider.spec.tsx
+++ b/__tests__/authkit-provider.spec.tsx
@@ -220,7 +220,6 @@ describe('useAuth', () => {
       entitlements: ['feature1'],
       impersonator: { email: 'admin@example.com' },
       oauthTokens: { access_token: 'token123' },
-      accessToken: 'access123',
     });
 
     const TestComponent = () => {
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -11,6 +11,7 @@ import { getWorkOS } from './workos.js';
  * @returns The sanitized auth object
  */
 function sanitize<T extends { accessToken?: string }>(value: T) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { accessToken, ...sanitized } = value;
   return sanitized;
 }
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -1,6 +1,7 @@
 'use server';
 
 import { signOut } from './auth.js';
+import { NoUserInfo, UserInfo } from './interfaces.js';
 import { refreshSession, withAuth } from './session.js';
 import { getWorkOS } from './workos.js';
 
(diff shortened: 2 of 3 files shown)


__tests__/authkit-provider.spec.tsx
```diff
@@ -220,7 +220,6 @@ describe('useAuth', () => {
       entitlements: ['feature1'],
       impersonator: { email: 'admin@example.com' },
       oauthTokens: { access_token: 'token123' },
```
cmatheson, line 222: we should probably remove these oauthTokens too for good measure (maybe someone else on the team thinks differently?).

nicknisi: It looks like that's only referenced in that test and isn't actually passed to the client. We're only sending `UserInfo | NoUserInfo` from these actions https://github.com/workos/authkit-nextjs/blob/700f29d3b1521ba35fafb1ab0f556992f1f9c80e/src/interfaces.ts#L26-L45

nicknisi: I went ahead and removed `oauthTokens` here from the mock return.

cmatheson: oh i see, we pass the oauthTokens to an onSuccess callback but otherwise don't store them at all :+1:

cmatheson (approved):

statico:
Hi there! How does this affect the `<UsersManagement />` widget as described in the [WorkOS Widgets docs](https://workos.com/docs/user-management/widgets/user-management)? Those docs say to use `getAccessToken` from `useAuth()`. (cc @tburgin)

cmatheson:
@statico `getAccessToken()` is for authkit-react. Nextjs falls into the backend integrations bucket ([token docs](https://workos.com/docs/user-management/widgets/integrating/tokens))

CleanShot 2025-03-13 at 14 50 36@2x

PR #214, opened 2025-03-13
add switchToOrganization action and client helper

nicknisi:
# Add Organization Switching Support

This PR adds a `switchToOrganization` function to AuthKit for NextJS, enabling users to switch between organizations with proper page revalidation.

## Features

- Server-side organization switching with flexible revalidation strategies
- Client-side support via the `useAuth` hook
- Support for path and tag-based revalidation

### Server Component 

```tsx
// In a server component or server action
import { switchToOrganization } from '@workos-inc/authkit-nextjs';

// Switch with default behavior (path revalidation)
await switchToOrganization('org_123');

// With custom options
await switchToOrganization('org_123', {
  returnTo: '/dashboard',
  revalidationStrategy: 'tag',
  revalidationTags: ['org-data']
});
```

### Client Component

```tsx
// In a client component
import { useAuth } from '@workos-inc/authkit-nextjs/components';

function OrganizationSwitcher() {
  const { organizationId, switchToOrganization } = useAuth();
  
  return (
    <select 
      value={organizationId} 
      onChange={(e) => switchToOrganization(e.target.value)}
    >
      {/* Organization options */}
    </select>
  );
}
```

This feature complements the Organization Switcher widget and provides a standardized approach for handling organization switching in NextJS applications.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -43,7 +43,7 @@ export async function switchToOrganization(
   const { returnTo, revalidationStrategy = 'path', revalidationTags = [] } = options;
   const headersList = await headers();
   const pathname = returnTo || headersList.get('x-url') || '/';
-  const result = refreshSession({ organizationId, ensureSignedIn: true });
+  const result = await refreshSession({ organizationId, ensureSignedIn: true });
 
   switch (revalidationStrategy) {
     case 'path':
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,12 +1,12 @@
 'use server';
 
-import { getAuthorizationUrl } from './get-authorization-url.js';
-import { cookies, headers } from 'next/headers';
-import { refreshSession, terminateSession } from './session.js';
-import { WORKOS_COOKIE_NAME, WORKOS_COOKIE_DOMAIN } from './env-variables.js';
 import { revalidatePath, revalidateTag } from 'next/cache';
+import { cookies, headers } from 'next/headers';
 import { redirect } from 'next/navigation';
-import { UserInfo, SwitchToOrganizationOptions } from './interfaces.js';
+import { WORKOS_COOKIE_DOMAIN, WORKOS_COOKIE_NAME } from './env-variables.js';
+import { getAuthorizationUrl } from './get-authorization-url.js';
+import { SwitchToOrganizationOptions, UserInfo } from './interfaces.js';
+import { refreshSession, terminateSession } from './session.js';
 
 export async function getSignInUrl({
   organizationId,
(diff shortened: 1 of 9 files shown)


src/auth.ts
```diff
@@ -25,4 +28,28 @@ async function signOut({ returnTo }: { returnTo?: string } = {}) {
...
 }
 
-export { getSignInUrl, getSignUpUrl, signOut };
+export async function switchToOrganization(
+  organizationId: string,
+  options: SwitchToOrganizationOptions = {},
+): Promise<UserInfo> {
+  const { returnTo, revalidationStrategy = 'path', revalidationTags = [] } = options;
+  const headersList = await headers();
+  const pathname = returnTo || headersList.get('x-url') || '/';
```
dcwither, line 37: 💯

src/auth.ts
```diff
@@ -25,4 +28,28 @@ async function signOut({ returnTo }: { returnTo?: string } = {}) {
...
+    case 'path':
+      revalidatePath(pathname);
+      break;
+    case 'tag':
+      for (const tag of revalidationTags) {
+        revalidateTag(tag);
+      }
+      break;
+  }
+  if (revalidationStrategy !== 'none') {
```
dcwither, line 50: I'd have to test this piece to ensure the other creations of auth tokens with widgets don't need both (not that the developer can't add this to their wrapper function).

src/auth.ts
```diff
@@ -25,4 +28,28 @@ async function signOut({ returnTo }: { returnTo?: string } = {}) {
...
-export { getSignInUrl, getSignUpUrl, signOut };
+export async function switchToOrganization(
+  organizationId: string,
+  options: SwitchToOrganizationOptions = {},
+): Promise<UserInfo> {
+  const { returnTo, revalidationStrategy = 'path', revalidationTags = [] } = options;
+  const headersList = await headers();
+  const pathname = returnTo || headersList.get('x-url') || '/';
+  const result = await refreshSession({ organizationId, ensureSignedIn: true });
+
```
dcwither, line 39: I think this chunk is something we would want to handle here
```ts
      const args = {
        organizationId,
        clientId: process.env.WORKOS_CLIENT_ID!, // probably have this within this package already
        provider: "authkit", // ?
        redirectUri: pathname,
      };

      if (err.error === "sso_required" || err.error === "mfa_enrollment") {
        const url = workos.userManagement.getAuthorizationUrl({
          ...args,
        });
        redirect(url);
      }
```

It ensures that we send them to the correct re-auth when switching fails. cc: @alisherry who has the right intersection of knowledge to discuss more clearly.

nicknisi: Looking at refreshSession would it ever throw an error with `err.error` equaling one of the above?

https://github.com/workos/authkit-nextjs/blob/c6de3414418c486f561264668ad0f2fc53826953/src/session.ts#L254-L325

nicknisi: @alisherry Thanks, I updated the PR with this error checking.

alisherry (approved):
Code LGTM, did not get a chance to test this. Thank you!

PR #218, opened 2025-03-17
add WORKOS_COOKIE_SAMESITE optional configuration

### LINKED ISSUE #216: Expose WORKOS_COOKIE_SAME_SITE option
We embed our application in Microsoft Teams as a [Tab application which requires cookies' SameSite to be set to 'None'](https://learn.microsoft.com/en-us/microsoftteams/platform/resources/samesite-cookie-update#tabs-dialogs-and-message-extensions)

nicknisi:
Add option to configure the [samesite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#controlling_third-party_cookies_with_samesite) cookie setting for specific use cases.

capture_20250317_115039

Fixes #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ WORKOS_COOKIE_NAME='authkit-cookie'
 WORKOS_API_HOSTNAME='api.workos.com' # base WorkOS API URL
 WORKOS_API_HTTPS=true # whether to use HTTPS in API calls
 WORKOS_API_PORT=3000 # port to use for API calls
+
+# Only change this if you specifically need cross-origin cookie support.
+WORKOS_COOKIE_SAMESITE='lax' # SameSite attribute for cookies: 'lax' (default), 'strict', or 'none'.
+
+>[!IMPORTANT] Security Note:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -54,11 +54,16 @@ WORKOS_COOKIE_NAME='authkit-cookie'
...
+
+# Only change this if you specifically need cross-origin cookie support.
+WORKOS_COOKIE_SAMESITE='lax' # SameSite attribute for cookies: 'lax' (default), 'strict', or 'none'.
 ```
 
-`WORKOS_COOKIE_DOMAIN` can be used to share WorkOS sessions between apps/domains.
-Note: The `WORKOS_COOKIE_PASSWORD` would need to be the same across apps/domains.
-Not needed for most use cases.
+>[!WARNING]
+>Setting `WORKOS_COOKIE_SAMESITE='none'` allows cookies to be sent in cross-origin contexts (like iframes), but reduces protection against CSRF attacks. This setting forces cookies to be secure (HTTPS only) and should only be used when absolutely necessary for your application architecture.
```
faroceann, line 63: 👍 nice

faroceann (approved):

PR #223, opened 2025-03-18
Expose a saveSession method for advanced use cases

nicknisi:
This PR adds and exposes a new `saveSession` method for advanced end-user cases. It also replaces duplicated code internally with calls to this.

---

While the standard authentication flow handles session management automatically, some use cases require manually creating and storing a session. This is useful for custom authentication flows like email verification or token exchange.

For these scenarios, you can use the `saveSession` function:

```typescript
import { saveSession } from '@workos-inc/authkit-nextjs';
import { getWorkOS } from '@workos-inc/authkit-nextjs';

// Example: Email verification flow
async function handleEmailVerification(req) {
  const { code } = await req.json();

  // Authenticate with the WorkOS API directly
  const authResponse = await getWorkOS().userManagement.authenticateWithEmailVerification({
    clientId: process.env.WORKOS_CLIENT_ID,
    code,
  });

  // Save the session data to a cookie
  await saveSession({
    accessToken: authResponse.accessToken,
    refreshToken: authResponse.refreshToken,
    user: authResponse.user,
    impersonator: authResponse.impersonator
  }, req);

  return Response.redirect('/dashboard');
}
```

>[!NOTE]
>This is an advanced API intended for specific integration scenarios. Most applications should use the standard
>authentication flow.

The `saveSession` function accepts either a `NextRequest` object or a URL string as its second parameter.

```typescript
// With NextRequest
await saveSession(session, req);

// With URL string
await saveSession(session, 'https://example.com/callback');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,8 +455,7 @@ async function handleEmailVerification(req) {
 ```
 
 >[!NOTE]
->This is an advanced API intended for specific integration scenarios. Most applications should use the standard
->authentication flow.
+>This is an advanced API intended for specific integration scenarios, such as those users using self-hosted AuthKit. If you're using hosted AuthKit you should not need this.
 
 The `saveSession` function accepts either a `NextRequest` object or a URL string as its second parameter.
 
--- a/README.md
+++ b/README.md
@@ -422,6 +422,52 @@ const organizations = await workos.organizations.listOrganizations({
 });
 ````
 
+### Advanced: Custom authentication flows
+
+While the standard authentication flow handles session management automatically, some use cases require manually creating and storing a session. This is useful for custom authentication flows like email verification or token exchange.
+
+For these scenarios, you can use the `saveSession` function:
+
+```typescript
+import { saveSession } from '@workos-inc/authkit-nextjs';
+import { getWorkOS } from '@workos-inc/authkit-nextjs';
+
+// Example: Email verification flow
+async function handleEmailVerification(req) {
+  const { code } = await req.json();
+
+  // Authenticate with the WorkOS API directly
+  const authResponse = await getWorkOS().userManagement.authenticateWithEmailVerification({
+    clientId: process.env.WORKOS_CLIENT_ID,
+    code,
+  });
+
+  // Save the session data to a cookie
+  await saveSession({
+    accessToken: authResponse.accessToken,
+    refreshToken: authResponse.refreshToken,
+    user: authResponse.user,
+    impersonator: authResponse.impersonator
+  }, req);
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -422,6 +422,52 @@ const organizations = await workos.organizations.listOrganizations({
...
+    user: authResponse.user,
+    impersonator: authResponse.impersonator
+  }, req);
+
+  return Response.redirect('/dashboard');
+}
+```
+
+>[!NOTE]
+>This is an advanced API intended for specific integration scenarios. Most applications should use the standard
```
cmatheson, line 458: i think if we wanted to be more specific, this method is useful for self-hosted authkit customers (hosted-authkit users shouldn't ever need to interface with this).

src/authkit-callback-route.ts
```diff
@@ -67,13 +65,7 @@ export function handleAuth(options: HandleAuthOptions = {}) {
...
         }
 
-        // The refreshToken should never be accesible publicly, hence why we encrypt it in the cookie session
-        // Alternatively you could persist the refresh token in a backend database
-        const session = await encryptSession({ accessToken, refreshToken, user, impersonator });
-        const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
-        const nextCookies = await cookies();
-
-        nextCookies.set(cookieName, session, getCookieOptions(request.url));
+        await saveSession({ accessToken, refreshToken, user, impersonator }, request);
```
cmatheson, line 68: this looks a lot nicer

src/session.ts
```diff
@@ -462,4 +451,45 @@ function getScreenHint(signUpPaths: string[] | undefined, pathname: string) {
...
+ *     user: authResponse.user,
+ *     impersonator: authResponse.impersonator
+ *   }, req);
+ * }
+ *
+ * @example
+ * // With a URL string
+ * await saveSession(session, 'https://example.com/callback');
+ */
+export async function saveSession(session: Session, request: NextRequest | string): Promise<void> {
```
cmatheson, line 487: this looks great, i wonder if saveSession should take an `AuthenticationResponse` instead of a `Session` (they're probably the same shape, but i think anytime you'd call this you just want to feed it the response of `authenticateWithFoo` ?

nicknisi: Yeah, looks like they're roughly the same shape. Thinking of changing the signature to this just for self-documenting:

```ts
export async function saveSession(
  sessionOrResponse: Session | AuthenticationResponse,
  request: NextRequest | string,
): Promise<void>;
```

cmatheson (approved):

PR #232, opened 2025-03-30
Respect sameSite while deleting cookie

mintuhouse:
Remove quotes around SameSite string


nicknisi (approved):
Good catch, and thanks for the contribution!

mintuhouse:
@nicknisi can you release a new version if possible?

PR #243, opened 2025-04-14
Fix errors thrown by signOut method

### LINKED ISSUE #242: signout() throws error
**Describe the bug**
getting error. using latest version of authkit sdk and nextjs.

```
 ⨯ TypeError: Invalid URL
    at getCookieOptions (../../src/cookie.ts:35:14)
    at signOut (../../src/auth.ts:28:42)
    at logoutAction (src/app/auth/actions.ts:17:16)
  15 |     console.log("no session id");
  16 |   }
> 17 |   await signOut({ returnTo: 'https://astradot.com' });
     |                ^
  18 | }
  19 |
  20 | export async function addOrg(org: string) { {
  code: 'ERR_INVALID_URL',
  input: '',
  digest: '6214018'
}

```

**To Reproduce**
```
export async function logoutAction(){
  await signOut({ returnTo: 'https://astradot.com' });
}
```
**Expected behavior**

**Screenshots**

Image

Image

 - authkit-nextjs version [e.g. 0.12.0]
 - Next.js version [e.g. 14.2.5]

**Additional context**

nicknisi:
Improve robustness of `signOut` method by:

- not calling `getCookieOptions` (the only options that may be needed are `path` and `domain`)
- remove call to `terminateSession` which would call `withAuth` after the cookie is deleted
- remove `terminateSession` method as it's no longer called by anything
- tests

Fixes #242

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -3,11 +3,11 @@
 import { revalidatePath, revalidateTag } from 'next/cache';
 import { cookies, headers } from 'next/headers';
 import { redirect } from 'next/navigation';
-import { WORKOS_COOKIE_NAME } from './env-variables.js';
+import { WORKOS_COOKIE_DOMAIN, WORKOS_COOKIE_NAME } from './env-variables.js';
 import { getAuthorizationUrl } from './get-authorization-url.js';
 import { SwitchToOrganizationOptions, UserInfo } from './interfaces.js';
-import { refreshSession, terminateSession } from './session.js';
-import { getCookieOptions } from './cookie.js';
+import { refreshSession, withAuth } from './session.js';
+import { getWorkOS } from './workos.js';
 export async function getSignInUrl({
   organizationId,
   loginHint,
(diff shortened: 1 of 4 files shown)


src/auth.ts
```diff
@@ -24,12 +24,29 @@ export async function getSignUpUrl({
...
-  await terminateSession({ returnTo });
+  let sessionId: string | undefined;
+
+  try {
+    const { sessionId: sid } = await withAuth();
+    sessionId = sid;
+  } finally {
+    const nextCookies = await cookies();
+    const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
+    const domain = WORKOS_COOKIE_DOMAIN || /* istanbul ignore next */ undefined;
```
ericroberts, line 41: Why do we need to ignore this?

ericroberts: oh it's test coverage, I didn't know what istanbul was. Thanks

mintuhouse:
@nicknisi This PR reverts the changes made in https://github.com/workos/authkit-nextjs/pull/232 which uses `getCookieOptions` even when deleting a cookie

nicknisi:
@mintuhouse Thata's correct because it lead to scenarios in #242. But, we're still calling `nextCookies.delete` with the same `name`, `domain`, and `path`. Are you running into issues because of this?

mintuhouse:
@nicknisi yes
the logout in MSTeams/iframe is broken now since sameSite setting is not being respected
Possible to use `getCookieOptions` in signOut too?

PR #247, opened 2025-04-29
fix: infinite redirects when using useAccessToken with 60 second token duration

nicknisi:
This change fixes an issue where the `useAccessToken` hook would cause infinite requests to the server when using tokens with a 60-second duration.

The problem:
- The token expiry buffer was set to 60 seconds
- When token lifetime matched this buffer exactly (60 seconds)
- Calculated delay was (expiry - buffer) = 0 seconds
- This triggered immediate refresh attempts in a loop

The fix:
- Added a minimum refresh delay of 15 seconds
- This prevents rapid refresh cycles even when calculations result in zero
- Applied in both updateToken and refresh functions
- Provides protection against token duration changes or clock issues

### DIFF
--- a/src/components/useAccessToken.ts
+++ b/src/components/useAccessToken.ts
@@ -3,8 +3,8 @@ import { getAccessTokenAction, refreshAccessTokenAction } from '../actions.js';
 import { useAuth } from './authkit-provider.js';
 
 const TOKEN_EXPIRY_BUFFER_SECONDS = 60;
-const MIN_REFRESH_DELAY = 15_000; // minimum delay before refreshing token
-const RETRY_DELAY = 5 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


src/components/useAccessToken.ts
```diff
@@ -3,6 +3,7 @@ import { getAccessTokenAction, refreshAccessTokenAction } from '../actions.js';
 import { useAuth } from './authkit-provider.js';
 
 const TOKEN_EXPIRY_BUFFER_SECONDS = 60;
+const MIN_REFRESH_DELAY = 15_000; // minimum delay before refreshing token
```
nholden, line 6: It's a little funky that we have these constants using different units. Might be helpful to pick one, or at least be more explicit about what units we're using in the name.

nholden (approved):
Thanks for the helpful explanation of the problem! Makes sense to me. Left a piece of non-blocking feedback.

PR #255, opened 2025-05-25
Respect sameSite while deleting cookie

mintuhouse:
Fix for logout when using from inside an iframe

Fixes https://github.com/workos/authkit-nextjs/pull/243#issuecomment-2878517485 
Redo of https://github.com/workos/authkit-nextjs/pull/232

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,8 +1,8 @@
 'use server';
 
-import { revalidatePath, revalidateTag } from 'next/cache';
-import { cookies, headers } from 'next/headers';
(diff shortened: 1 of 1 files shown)


mintuhouse:
@nicknisi can you review this? 
we need a release with fix for this to use the new org switching helper

nicknisi:
@mintuhouse I did some cleanup on this and some work on the `getCookieOptions` to make it compatible with calling in `signOut` to ensure the cookie options are the same. I continued this work in #281. Thanks!

PR #257, opened 2025-05-28
feat: handleAuth > return authenticationMethod

dlarroder:
This change just exposes the `authenticationMethod` used by the user in handleAuth

We have a use case where we need to show the authentication method of the user from our UI, this change would make it easier for us to sync our data with WorkOS.

We have submitted a feature request for this in slack linked [here](https://aphex-group.slack.com/archives/C0793738B6V/p1747795602716859).


nicknisi:
Thanks! I'll follow up with a fast follow to add a section about it to the README. I appreciate the contribution!

PR #292, opened 2025-08-18
add prompt to getAuthorizationURL

jameslcarpino:
updated getAuthorizationURL to accept consent = 'prompt' to force user to consent for oauth. This is so that you can expect to get a refresh token with oauth

### DIFF
--- a/__tests__/auth.spec.ts
+++ b/__tests__/auth.spec.ts
@@ -78,8 +78,13 @@ describe('auth.ts', () => {
     });
   });
 
-  it('should use prompt=consent when calling getSignInUrl', async () => {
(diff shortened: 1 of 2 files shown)


src/get-authorization-url.ts
```diff
@@ -21,6 +22,7 @@ async function getAuthorizationUrl(options: GetAuthURLOptions = {}) {
     screenHint,
     organizationId,
     loginHint,
+    prompt,
```
nicknisi, line 25: This is failing because this library is on an older version of `@workos-inc/node`.

```json
"@workos-inc/node": "^7.37.1",
```

nicknisi (changes requested):

nicknisi (approved):

PR #294, opened 2025-08-25
Add afterAuth to `authkitMiddleware` function

parkersm1th-magical:
Add afterAuth to `authkitMiddleware` so folks don't have to compose the middleware themselves but can run code after auth, before .next


parkersm1th-magical:
cc @nicknisi

PR #311, opened 2025-09-24
Allow onSuccess callback to update session

nholden:
Today, it's not possible to refresh a token and save it to the session in the `onSuccess` callback in your callback route. That's because `handleAuth` calls `saveSession` with the original token after it calls your `onSuccess` callback.

This PR allows changes you save to the session in your `onSuccess` callback to persist after redirecting.


nholden:
> When shipping this, would you consider this to be a minor bump or a patch?

Minor bump makes sense for the reason you mentioned!

PR #321, opened 2025-10-14
feat: allow dynamic redirect on success

axlEscalada:
This change allow to dynamic redirect using `onSuccess` return, only `NextResponse` is allowed here.
This solves cases when we want to redirect based on `HandleAuthSuccessData` (e.g. Vercel integration where they send query params)

### DIFF
--- a/src/authkit-callback-route.spec.ts
+++ b/src/authkit-callback-route.spec.ts
@@ -284,7 +284,7 @@ describe('authkit-callback-route', () => {
 
       const handler = handleAuth({
         onSuccess: async (data) => {
-            return NextResponse.redirect('https://example.com/dashboard');
+          return NextResponse.redirect('https://example.com/dashboard');
         },
       });
 


src/authkit-callback-route.ts
```diff
@@ -96,7 +96,7 @@ export function handleAuth(options: HandleAuthOptions = {}) {
         await saveSession({ accessToken, refreshToken, user, impersonator }, request);
 
         if (onSuccess) {
-          await onSuccess({
+          const onSuccessResponse = await onSuccess({
```
nicknisi: suggestion: I wonder if it makes sense to provide the default response as an argument to onSuccess? That way, you don't always need to create a new one from scratch.

This would also help make it more explicit that if `onSuccess` returns a `NextResponse`, it's going to take precedence over `returnPathname` if both are provided.

axlEscalada:
> Thanks for the contribution! The functionality is useful. I have a few concerns:
> 
> 1. **Naming doesn't signal control flow**: The name `onSuccess` typically implies side-effects only (logging, analytics, etc.), not that the callback can control the redirect. However, since `onSuccess` is already public API, renaming would be a breaking change, so adding clear documentation is probably the better option here.
> 2. **Ambiguity when `onSuccess` returns a response and `returnPathname` is defined**: I think this can be somewhat alleviated by providing the `response` to `onSuccess`, helping to imply that if `onSuccess` returns a response, that will take precedence. Would you be open to adding that?
> 
> Would love to hear your thoughts. Thanks!

1. I agree with this. I considered adding a new callback function named onSuccessReturnTo but that approach would introduce more drawbacks mainly having to define which one takes precedence over the other. I can document this new behavior in the README for clarity
2. I think this is a good approach passing a default response, however if we change the signature to:
` onSuccess?: (data: HandleAuthSuccessData, defaultResponse: Response) => void | NextResponse | Promise<void> | Promise<NextResponse>;` it would be breaking change right? making it optional `defaultResponse?: Response` feels a bit odd, as a user I would wonder why defaultResponse might be optional.

nicknisi:
> I think this is a good approach passing a default response, however if we change the signature to:
 onSuccess?: (data: HandleAuthSuccessData, defaultResponse: Response) => void | NextResponse | Promise<void> | Promise<NextResponse>; it would be breaking change right? making it optional defaultResponse?: Response feels a bit odd, as a user I would wonder why defaultResponse might be optional.

Yeah, I we could add it to the `HandleAuthSuccessData` interface and avoid having a second argument. What do you think?

PR #323, opened 2025-10-19
feat: Add initialAuth to AuthkitProvider

### LINKED ISSUE #295: AuthKitProvider - redundant api call
AuthKitProvider use mount effect to call getAuth and get user login, we already know who is the current user in ssr and can pass it via server component to improve speed and remove redundant call

danielr18:
Fixes #295

`AuthKitProvider` was always fetching auth data on mount. Added an optional `initialAuth` prop so you can pass the data directly from your server component and skip the unnecessary round trip.

**Example:**
```tsx
export default async function RootLayout({ children }) {
  const auth = await withAuth();
  const { accessToken, ...initialAuth } = auth;

  return (
    <AuthKitProvider initialAuth={initialAuth}>{children}</AuthKitProvider>
  );
}
```

Not sure if this is the best approach, but there should be a way to avoid the server action call on every initial mount if preferred.

### DIFF
--- a/src/components/authkit-provider.tsx
+++ b/src/components/authkit-provider.tsx
@@ -43,7 +43,7 @@ interface AuthKitProviderProps {
   /**
    * Initial auth data from the server. If provided, the provider will skip the initial client-side fetch.
    */
-  initialAuth?: Omit<UserInfo | NoUserInfo, "accessToken">;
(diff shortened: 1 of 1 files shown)


nicknisi (approved):
This looks good, thank you! I pushed up a few commits to fix tests and run formatting.

PR #332, opened 2025-11-02
fix: prevent caching issues on authenticated pages

nicknisi:
## Summary

Fixes a potential session crossover vulnerability in CDN environments (particularly CloudFront with SST/OpenNext deployments).

Without `Vary: Cookie`, CDNs can't differentiate between User A and User B, potentially serving User A's cached authenticated content to User B.

## Changes

- Added cache security headers to middleware responses when authentication is detected
- Added cache security headers to OAuth callback routes
- Updated README with CDN deployment guidance

**Headers now set on authenticated requests:**
- `Cache-Control: private, no-store, must-revalidate`
- `Vary: Cookie` (critical - tells CDNs to differentiate users)
- `CDN-Cache-Control: no-store` (CloudFront/Vercel specific)
- `x-middleware-cache: no-cache` (OpenNext/SST specific)
- `Pragma: no-cache` (HTTP/1.0 compatibility)

**Detection logic:**
- Active session with access token
- Session cookie present in request
- Authorization header present

Auth callback routes are always protected via direct function call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -794,9 +794,8 @@ AuthKit automatically implements cache security measures to protect against sess
 The library automatically sets appropriate cache headers on all authenticated requests:
 
 - `Cache-Control: private, no-store, must-revalidate` - Prevents CDN caching of authenticated responses
-- `Vary: Cookie` - Ensures CDNs differentiate between different users
-- `CDN-Cache-Control: no-store` - Additional protection for CloudFront and Vercel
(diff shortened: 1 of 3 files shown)


nicknisi:
> The overall approach seems good. Are there any of the server-action stuff (switchOrganization) that need similar treatment?

Correct me if I'm wrong but I think those should all be fine.

PR #335, opened 2025-12-02
fix: clear session cookies when onSuccess fails

### LINKED ISSUE #334: Authentication cookies are set even when onSuccess callback throws an error
**Describe the bug**
The current behavior updates authentication cookies before executing the onSuccess callback. If the onSuccess function throws an error, the callback route correctly returns an error, but the cookies are still set. This makes the authentication appear successful even though the success logic failed.

**To Reproduce**
1. Add an onSuccess function inside handleAuth (in app/callback/route.ts) (for next-authkit-example)
2. Inside onSuccess, throw an error.
3. Observe that the callback route returns an error, but the authentication cookie is still set.

**Expected behavior**
The callback route should throw an error (already happening), and the cookies should not be set when onSuccess fails.

**Screenshots**

- OS: Windows
- Browser: Chrome
- authkit-nextjs version: 2.1.0
- Next.js version: 15.2.3

**Additional context**
Potential solution: If the onSuccess function throws an error, cookies should not be set.

LShaViR:
Fixes #334

**Overview**
This pull request fixes an issue where authentication cookies were being set even when the onSuccess callback inside handleAuth threw an error. This resulted in inconsistent auth state: the callback route returned an error, but the user was still marked as authenticated.

**What’s changed**

- Updated the authentication flow to ensure cookies are only set after the onSuccess callback completes successfully.

- If onSuccess throws an error, cookie-setting is skipped and the error is correctly propagated to the callback route.

**Reason for the change**
The previous behavior caused misleading authentication states and made debugging difficult. Authentication should not be considered successful if onSuccess fails.

**How to test**

1. Add an onSuccess function in app/callback/route.ts.
2. Throw an error inside it.
3. Confirm that:  
  - The callback route shows the error.  
  - No cookies are set.

**Impact**
Ensures correctness and consistency in the authentication flow and prevents unintended authenticated sessions.

### DIFF
--- a/src/authkit-callback-route.spec.ts
+++ b/src/authkit-callback-route.spec.ts
@@ -51,7 +51,7 @@ describe('authkit-callback-route', () => {
 
     beforeAll(() => {
       // Silence console.error during tests
-      jest.spyOn(console, 'error').mockImplementation(() => {});
+      jest.spyOn(console, 'error').mockImplementation(() => { });
     });
 
     beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


LShaViR:
@nicknisi 
That would be better. And Instead of deleting the session, what if we delete only the cookies so that other applications on the same platform (e.g., youtube, drive , docs ... for google) continue to work without any interruption?

I’ve made some changes for the delete-cookies solution, and it passes all tests as well. ( should i commit it in same pull request)

nicknisi:
Updated this branch against current main and resolved the callback conflicts. Failed onSuccess callbacks now await session cleanup using the configured cookie name and scope. Session reads/updates during successful callbacks and the existing redirect-rethrow behavior remain supported.

Regression tests use real Next response-cookie serialization to check cleanup ordering and attributes rather than only the Map mock. Verification: 417 tests passed, including coverage, plus lint, typecheck, formatting, and build. A fresh independent code review found no blocker. Not merged or released.

nicknisi (approved):

PR #340, opened 2025-12-10
fix: handle full URLs in returnPathname to prevent malformed redirects

nicknisi:
## Summary

- Fixes an edge case where `returnPathname` containing a full URL (e.g., `https://example.com/path`) would produce malformed redirect URLs like `https://app.com/https://example.com/path`
- Normalizes `returnPathname` through URL parsing to always extract just the pathname portion
- Simplifies the code by removing the conditional branch while making it more robust


cmatheson (approved):
i think it's kind of surprising that we just ignore the host portion of the url ... should we return an error if the state-host doesn't match the request host? probably not too relevant in practice.

nicknisi:
> i think it's kind of surprising that we just ignore the host portion of the url ... should we return an error if the state-host doesn't match the request host? probably not too relevant in practice.

@cmatheson I'll consider this. I think changing that could be considered a breaking change so I'll want to do it in a separate PR/release.

PR #347, opened 2025-12-19
feat: enhance onSuccess callback and middleware request

abedshaaban:
This feature introduces a main change:
- improved `authkitMiddleware` to be able to merge multiple middlewares in Nextjs and be able to modify them.

Usage examples:

### For middlewares
```ts
import { authkitMiddleware } from '@workos-inc/authkit-nextjs'

export default authkitMiddleware(async (auth, request, event) => {
  const session = await auth()
  
  if (isProtectedRoute(request) && !session.user) {
    // Handle protection logic
  }
  
  // Chain with other middleware (e.g., i18n)
  return handleI18nRouting(request)
})
```

### DIFF
--- a/src/authkit-callback-route.spec.ts
+++ b/src/authkit-callback-route.spec.ts
@@ -270,7 +270,6 @@ describe('authkit-callback-route', () => {
       await handler(request);
 
       expect(onSuccess).toHaveBeenCalledWith(mockAuthResponse);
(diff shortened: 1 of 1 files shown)


abedshaaban:
Not 100% sure but I tried to implement it as clerk does in their middleware to combine middlewares.

PR #349, opened 2025-12-23
feat: add TokenRefreshError with userId and sessionId for debugging

nicknisi:
## Problem

When token refresh fails, the error doesn't include enough context to debug which user/session was affected. This makes it harder to correlate errors with specific users in logs or the WorkOS dashboard.

## Solution

Add `TokenRefreshError` (matching the pattern in `authkit-session`) that includes `userId` and `sessionId` on the error object. These are extracted from the session before throwing.

## Example Output

```
⨯ Error [TokenRefreshError]: Failed to refresh session: Error: invalid_grant
Error Description: Session has already ended.
    at async testRefresh (src/app/test-error/actions.ts:7:5)
   5 | export async function testRefresh() {
   6 |   try {
>  7 |     await refreshSession();
     |     ^
   8 |     return { success: true };
   9 |   } catch (error) {
  10 |     if (error instanceof TokenRefreshError) { {
  data: undefined,
  userId: 'user_01ABC123...',
  sessionId: 'session_01XYZ789...',
  digest: '1995551028',
  [cause]: OauthException: Error: invalid_grant
  Error Description: Session has already ended.
      at Generator.throw (<anonymous>) {
    status: 400,
    requestID: '',
    error: 'invalid_grant',
    errorDescription: 'Session has already ended.',
    rawData: {
      error: 'invalid_grant',
      error_description: 'Session has already ended.'
    }
  }
}
```

Consumers can catch and inspect these properties:

```typescript
import { TokenRefreshError } from '@workos-inc/authkit-nextjs';

try {
  await refreshSession();
} catch (error) {
  if (error instanceof TokenRefreshError) {
    console.error('Refresh failed for user:', error.userId, 'session:', error.sessionId);
  }
}
```

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,0 +1,46 @@
+import type { Session } from './interfaces.js';
+import { decodeJwt } from './jwt.js';
+
+export class AuthKitError extends Error {
+  data?: Record<string, unknown>;
+
+  constructor(message: string, cause?: unknown, data?: Record<string, unknown>) {
+    super(message);
+    this.name = 'AuthKitError';
+    this.cause = cause;
+    this.data = data;
+  }
+}
+
+export interface TokenRefreshErrorContext {
+  userId?: string;
+  sessionId?: string;
+}
+
+export class TokenRefreshError extends AuthKitError {
+  readonly userId?: string;
+  readonly sessionId?: string;
+
+  constructor(message: string, cause?: unknown, context?: TokenRefreshErrorContext) {
+    super(message, cause);
+    this.name = 'TokenRefreshError';
+    this.userId = context?.userId;
+    this.sessionId = context?.sessionId;
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/errors.ts
```diff
@@ -0,0 +1,55 @@
...
+  const context: TokenRefreshErrorContext = {};
+
+  if (session.user?.id) {
+    context.userId = session.user.id;
+  }
+
+  if (session.accessToken) {
+    try {
+      const { payload } = decodeJwt(session.accessToken);
+      if (payload.sid) {
```
cmatheson: nit: i think we should pull the user_id (`sub`) from the access token here too (instead of from the session above). i agree the session.user.id _should_ always be the same but since we already have the access token decoded here it makes sense to me to use that as the source of truth).

cmatheson (approved):

faroceann (approved):
lgtm

PR #352, opened 2025-12-28
Replace useState calls with discriminated union

sundaray:
This PR replaces 10 separate `useState` calls inside `authkit-provider.tsx` with a single discriminated union:

```ts
type AuthState =
 | {
     status: 'loading';
   }
 | { status: 'unauthenticated'; data: Omit<NoUserInfo, 'accessToken'> }
 | { status: 'authenticated'; data: Omit<UserInfo, 'accessToken'> };
```

This brings two benefits to the codebase:

 - Makes it impossible for anyone to unintentionally create an invalid auth state combination (which is possible now).
 - Makes it immediately clear which auth state is being handled at any point in the code. 

Note: There are no changes to the public API. All existing tests pass.

### DIFF
--- a/src/components/authkit-provider.tsx
+++ b/src/components/authkit-provider.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { createContext, ReactNode, useCallback, useContext, useEffect, useState } from 'react';
+import React, { createContext, ReactNode, useCallback, useContext, useEffect, useReducer } from 'react';
 import {
   checkSessionAction,
   getAuthAction,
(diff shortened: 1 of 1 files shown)


src/components/authkit-provider.tsx
```diff
@@ -46,43 +43,63 @@ interface AuthKitProviderProps {
...
-  const [sessionId, setSessionId] = useState<string | undefined>(initialAuth?.sessionId);
-  const [organizationId, setOrganizationId] = useState<string | undefined>(initialAuth?.organizationId);
-  const [role, setRole] = useState<string | undefined>(initialAuth?.role);
-  const [roles, setRoles] = useState<string[] | undefined>(initialAuth?.roles);
-  const [permissions, setPermissions] = useState<string[] | undefined>(initialAuth?.permissions);
-  const [entitlements, setEntitlements] = useState<string[] | undefined>(initialAuth?.entitlements);
-  const [featureFlags, setFeatureFlags] = useState<string[] | undefined>(initialAuth?.featureFlags);
-  const [impersonator, setImpersonator] = useState<Impersonator | undefined>(initialAuth?.impersonator);
-  const [loading, setLoading] = useState(!initialAuth);
+  const [authState, setAuthState] = useState<AuthState>(() => createAuthState(initialAuth));
```
nicknisi: I wonder if this should be a `useReducer` call instead? That way the auth state and loading boolean can be updated independently?

PR #401, opened 2026-04-08
fix: allow dashboard impersonation to bypass PKCE/CSRF verification

### LINKED ISSUE #400: Dashboard impersonation broken with v3 PKCE/CSRF verification
**Describe the bug**

After upgrading to v3.0.0, impersonating a user from the WorkOS dashboard fails with `Error: Missing required auth parameter`. The v3 PKCE/CSRF verification rejects the callback because dashboard impersonation bypasses the application's proxy/middleware, so the `wos-auth-verifier` cookie and `state` parameter are never set.

Normal sign-in flows work correctly.

**To Reproduce**

1. Set up `@workos-inc/authkit-nextjs` v3.0.0 with the composable proxy pattern (`authkit()` + `handleAuthkitProxy()`) and `handleAuth()` on the callback route
2. Sign in normally to confirm auth works ✅
3. Go to the WorkOS dashboard → Users → select a user → click "Impersonate"
4. WorkOS redirects to `/callback?code=...` (no `state` parameter)
5. See error: `[AuthKit callback error] Error: Missing required auth parameter` → 500

**Expected behavior**

Dashboard impersonation should complete successfully as it did in v2. The impersonation flow is initiated from WorkOS, not from the application, so the application has no opportunity to set the `wos-auth-verifier` cookie or include a `state` parameter.

For reference, a normal sign-in callback includes both:

```
GET /callback?code=01KNN...&state=Fe26.2*1*756e5... → 200
```

But impersonation only includes the code:

```
GET /callback?code=01KNNQAHQBT2FE4Z74QPGMZ9KT → 500
```

- Browser: Chrome
- authkit-nextjs version: 3.0.0
- Next.js version: 16 (using `proxy.js` with composable `authkit()` + `handleAuthkitProxy()` pattern)

**Additional context**

Setup details:
- Callback path is excluded from the proxy matcher (as recommended in the docs)
- Using WorkOS hosted sign-in pages

The root cause is that v3's PKCE/CSRF hardening assumes every callback was initiated by the application (where the proxy sets the cookie + state). Dashboard impersonation initiates from WorkOS itself, skipping that step entirely.

Thanks for the great work on v3, the composable proxy pattern is a really nice improvement! Happy to provide any additional info or test a fix if that would be helpful. 🙏

nicknisi:
## Summary

Fixes #400 — dashboard impersonation broken with v3 PKCE/CSRF verification.

- Dashboard impersonation is initiated by WorkOS (not the app), so the callback receives only `code` with no `state` param. The v3 PKCE hardening rejected these callbacks with `Missing required auth parameter`.
- When `state` is absent (`null`), skip PKCE/CSRF verification and exchange the code without a `codeVerifier`. After exchange, verify the response contains `impersonator` data — reject otherwise, preventing intercepted non-impersonation codes from being accepted on this path.
- Uses `state === null` (not `!state`) so empty-string state is still treated as a malformed normal flow, not impersonation.
- Normal sign-in flows with `state` still go through full PKCE/CSRF verification.

## Security considerations

- PKCE (RFC 7636) cannot be applied to server-initiated flows — the client never initiates the authorization request, so there's no opportunity to generate a code verifier.
- The `impersonator` check after code exchange is the compensating control. If the response doesn't indicate impersonation, the session is rejected.
- Attack chain for exploitation is narrow: intercept redirect to server endpoint → exchange before legitimate request (single-use code) → code must be an impersonation code (impersonator check).

## Test plan

- [x] `pnpm test` passes (345 tests)
- [x] `pnpm run build` succeeds
- [x] `pnpm run lint` passes
- [x] Impersonation succeeds with code only (no state, no cookie)
- [x] Impersonation succeeds with stale PKCE cookie present (real-world scenario)
- [x] Rejects when PKCE bypassed but response has no impersonator
- [x] Empty state string (`?state=`) is not treated as impersonation
- [x] Normal PKCE/CSRF flows still fully verified
- [x] Tested end-to-end in example app via WorkOS dashboard impersonation

### DIFF
--- a/src/authkit-callback-route.spec.ts
+++ b/src/authkit-callback-route.spec.ts
@@ -472,14 +472,29 @@ describe('authkit-callback-route', () => {
         expect(response.status).not.toBe(500);
       });
 
-      it('should return 500 when neither state nor cookie exist', async () => {
+      it('should still reject when state is present but cookie is missing', async () => {
         vi.mocked(workos.userManagement.authenticateWithCode).mockResolvedValue(mockAuthResponse);
 
         request.nextUrl.searchParams.set('code', 'test-code');
+        request.nextUrl.searchParams.set('state', 'some-state');
 
         const handler = handleAuth();
         const response = await handler(request);
 
+        expect(workos.userManagement.authenticateWithCode).not.toHaveBeenCalled();
+        expect(response.status).toBe(500);
+      });
+
+      it('should not treat empty state string as impersonation', async () => {
+        vi.mocked(workos.userManagement.authenticateWithCode).mockResolvedValue(mockAuthResponse);
+
+        request.nextUrl.searchParams.set('code', 'test-code');
+        request.nextUrl.searchParams.set('state', '');
+
+        const handler = handleAuth();
+        const response = await handler(request);
+
+        // Empty state is not null — should enter normal PKCE flow and fail (no matching cookie)
         expect(workos.userManagement.authenticateWithCode).not.toHaveBeenCalled();
         expect(response.status).toBe(500);
       });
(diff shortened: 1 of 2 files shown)


gjtorikian (approved):

cmatheson (changes requested):
we shouldn't ever bypass pkce/csrf verification. the solution to this is to set up an initiate_login_uri.

PR #402, opened 2026-04-09
docs: add sign-in endpoint setup and impersonation requirements

nicknisi:
## Summary

- Adds a **Sign-in endpoint** section to the README Setup guide with a route example and WorkOS dashboard configuration instructions
- Adds a note to the **Visualizing an impersonation** section that a sign-in endpoint is required
- Adds a **Troubleshooting** entry for the `Missing required auth parameter` error during dashboard impersonation

Context: #400 — dashboard impersonation fails with v3 because WorkOS redirects directly to the callback without `state`. The fix is to configure a sign-in endpoint (`initiate_login_uri`) so impersonation routes through the app first, allowing PKCE/state to be set up normally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The `onSuccess` callback receives the following data:
 
 ### Sign-in endpoint
 
-Create a route that initiates the AuthKit sign-in flow. This route is used as the **Sign-in endpoint** (also known as `initiate_login_uri`) in your WorkOS dashboard settings.
+Create a route that initiates the AuthKit sign-in flow. This route is used as the **[Sign-in endpoint](https://workos.com/docs/authkit/nextjs/2-configure-your-project/configure-a-redirect-uri#sign-in-endpoint)** (also known as `initiate_login_uri`) in your WorkOS dashboard settings.
 
 ```ts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,6 +142,26 @@ The `onSuccess` callback receives the following data:
 
 **Note**: `authenticationMethod` is only provided during the initial authentication callback. It will not be available in subsequent requests or session refreshes.
 
+### Sign-in endpoint
+
+Create a route that initiates the AuthKit sign-in flow. This route is used as the **Sign-in endpoint** (also known as `initiate_login_uri`) in your WorkOS dashboard settings.
```
gjtorikian: Is there a documentation URL we can link to here for folks who want/need more info?

nicknisi: Good call — linked it to https://workos.com/docs/authkit/nextjs/2-configure-your-project/configure-a-redirect-uri#sign-in-endpoint in 1259d51. That page has the Next.js-specific explanation of why the endpoint exists (handling bookmarked hosted pages, password reset email links, etc.) along with the same `getSignInUrl` pattern.

gjtorikian (approved):

PR #403, opened 2026-04-10
fix: isolate concurrent PKCE flows to prevent cookie clobbering

cn-stephen:
When a user's session expires (e.g. inactivity timeout) and they have multiple tabs open all tabs wake up and fire requests simultaneously, or even if a single tab fires off multiple requests at once that trigger the auth flow. Each request hits the middleware, which generates a fresh PKCE state and overwrites the single shared `wos-auth-verifier` cookie. The last write wins, so when callbacks return from WorkOS, every tab except the last one fails with "OAuth state mismatch" or "Auth cookie missing".

This is reliably reproducible: open 3-4 tabs, wait for session expiry, switch back to the browser. The tabs wake up concurrently, each triggers a middleware redirect with a different PKCE state, and the shared cookie gets clobbered.

Two changes fix this:

1. Per-flow cookie names: derive a unique suffix from an FNV-1a hash of the sealed state (e.g. `wos-auth-verifier-a3f7c012`). Each concurrent flow gets its own cookie, and each callback reads the correct one. Uses @sindresorhus/fnv1a for a fast, format-agnostic 32-bit hash.

2. Document-only cookies: reuse the existing `isInitialDocumentRequest` check to only set PKCE cookies on full page navigations. Fetch, XHR, RSC, and prefetch requests never follow cross-origin redirects so they can never complete the OAuth flow. Without this guard, the per-flow cookie names cause many cookies per tab (one per concurrent request e.g., react-query reissuing XHR fetches on window activate), which quickly exceeds browser header limits and triggers HTTP 431 (Request Header Fields Too Large).

### DIFF
--- a/src/cookie.ts
+++ b/src/cookie.ts
@@ -92,10 +92,13 @@ export function getCookieOptions(
   };
 }
 
+const PKCE_COOKIE_MAX_AGE = 600; // 10 minutes
+
 /**
  * Cookie options for the PKCE verifier cookie.
  * 'strict' blocks the cookie on the cross-site redirect back from WorkOS; downgrade to 'lax'.
  * 'none' is more permissive and must be preserved for iframe/cross-origin embed flows.
+ * Max-age is always capped to 10 minutes — PKCE cookies are single-use and short-lived.
  */
 export function getPKCECookieOptions(): CookieOptions;
 export function getPKCECookieOptions(redirectUri: string | null | undefined, asString: true, expired?: boolean): string;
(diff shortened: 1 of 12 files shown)


cn-stephen:
Tested locally, works great.

src/cookie.ts
```diff
@@ -111,13 +114,16 @@ export function getPKCECookieOptions(
 ): CookieOptions | string {
   if (asString) {
     const options = getCookieOptions(redirectUri, true, expired);
-    return options.replace(/SameSite=Strict/i, 'SameSite=Lax');
+    return options
+      .replace(/SameSite=Strict/i, 'SameSite=Lax')
+      .replace(/Max-Age=\d+/, `Max-Age=${expired ? 0 : PKCE_COOKIE_MAX_AGE}`);
```
cn-stephen, line 119: ngl this string replace technique is fragile as heck. But it existed and I'm not trying to clean up everything possible.

Hope this is OK for now

cn-stephen:
cc @nicknisi as well, as you're the only guy I know on the team xD and we had a good collab previously.

Hope the explanations are sufficient

cn-stephen:
bump. This has some important improvements I hope would be considered. @nicknisi @jonatascastro12

nicknisi (approved):
Thanks! Merging with two small additions:

1. Legacy cookie fallback in callback (handles v3.0.x users mid-upgrade)
2. signOut clears all `wos-auth-verifier*` cookies

PR #423, opened 2026-05-06
fix(deps): move @workos-inc/node to peerDependencies

### LINKED ISSUE #422: TS2883: inferred return types from withAuth() / refreshSession() are non-portable when consumer pins a different @workos-inc/node version
**Describe the bug**

TypeScript emits `TS2883` (and on some configurations, the related `TS2742`) on consumer code that infers return types from `withAuth()` or `refreshSession()`:

```
error TS2883: The inferred type of 'requireAuth' cannot be named without a reference to 'User' from
'@workos-inc/authkit-nextjs/node_modules/@workos-inc/node'. This is likely not portable.
A type annotation is necessary.
```

The error reproduces whenever the consumer's resolved version of `@workos-inc/node` differs from the version that satisfies authkit-nextjs's transitive `^9.0.0` range. Package managers then install a duplicate nested copy under `node_modules/@workos-inc/authkit-nextjs/node_modules/@workos-inc/node/`, and TypeScript can no longer write a portable declaration for any inferred type that surfaces `User` (or `OauthTokens`, `AuthenticationResponse`) through the public API.

**To Reproduce**

1. In a Next.js project, install `@workos-inc/authkit-nextjs@^4.0.1` and pin `@workos-inc/node` directly to a version that doesn't match `^9.0.0`'s currently-resolved point version, e.g. `9.2.0` while authkit-nextjs's transitive resolution is still at `9.1.1`. With Yarn 4 + node-modules linker this is what `dependabot[bot]` produces every time it bumps the direct pin.
2. Write any function whose return type is inferred from `withAuth()`. A minimal repro:

   ```ts
   // lib/auth/session.ts
   import { withAuth, getSignInUrl } from "@workos-inc/authkit-nextjs"
   import { redirect } from "next/navigation"

   export async function requireAuth() {
     const session = await withAuth()
     if (!session.user) {
       const signInUrl = await getSignInUrl()
       redirect(signInUrl)
     }
     return session as typeof session & { user: NonNullable<typeof session.user> }
   }
   ```

3. Run `tsc --noEmit` (or in any monorepo package with `declaration: true` / `composite: true` / `isolatedDeclarations`).
4. Observe `TS2883` on the `requireAuth` declaration.

Confirmed in `yarn.lock` that there are two resolved versions of `@workos-inc/node`:

```
"@workos-inc/node@npm:9.2.0":         # consumer direct dep
"@workos-inc/node@npm:^9.0.0":        # transitive from authkit-nextjs → 9.1.1
```

`yarn dedupe --check '@workos-inc/node'` reports the split:

```
@workos-inc/node@npm:^9.0.0 can be deduped from @workos-inc/node@npm:9.1.1 to @workos-inc/node@npm:9.2.0
```

**Expected behavior**

`tsc` should emit a portable declaration. authkit-nextjs's public types should resolve to a single canonical copy of `@workos-inc/node`'s types regardless of the consumer's resolved version, so that any inferred type referencing `User` / `UserInfo` / `Session` / `HandleAuthSuccessData` is namable from a stable path.

**Root cause**

`@workos-inc/node` is declared in `dependencies` (https://github.com/workos/authkit-nextjs/blob/main/package.json#L41), but its types are surfaced into authkit-nextjs's public-API interfaces in `src/interfaces.ts`:

```ts
import type { AuthenticationResponse, OauthTokens, User } from '@workos-inc/node';

export interface Session  { user: User; ... }
export interface UserInfo { user: User; ... }
export interface HandleAuthSuccessData extends Session {
  oauthTokens?: OauthTokens;
  authenticationMethod?: AuthenticationResponse['authenticationMethod'];
  ...
}
```

A regular `dependencies` entry on a SDK whose types appear in the public API allows the package manager to install a separate nested copy whenever the consumer's resolved version diverges. The same hazard exists with `next` and `react` — and authkit-nextjs already declares both as `peerDependencies` for exactly this reason.

**Proposed fix**

Move `@workos-inc/node` from `dependencies` to `peerDependencies`. The peer-dep contract forces a single resolved copy in the consumer tree by construction. PR coming.

This is a breaking change, but in practice the impact is small: every authkit-nextjs consumer already has to call `@workos-inc/node` directly for the parts of the WorkOS SDK that authkit doesn't wrap (organization management, user management, FGA, audit logs, etc.).

**Workaround (for anyone hitting this today)**

`yarn dedupe '@workos-inc/node'` in the consumer's repo collapses the duplicate, and either a `resolutions` pin or a CI-side `yarn dedupe --check '@workos-inc/node'` step (run after install) prevents the regression from recurring on future Dependabot bumps.

**Environment**

- `authkit-nextjs` version: 4.0.1 (also affects all 4.x; same class of bug applied to earlier majors with their own `@workos-inc/node` ranges)
- `@workos-inc/node` versions involved: 9.1.1 (transitive) vs 9.2.0 (consumer direct)
- Next.js version: any
- TypeScript version: 6.0.x (also reproduces on 5.x with the same diagnostic code)
- Package manager: Yarn 4.14.1 with `nodeLinker: node-modules` (the same hazard exists on npm/pnpm whenever versions split)
- OS: macOS / Linux (build-time only — not runtime, not browser-dependent)

**Additional context**

Same root cause manifested as `TS2742` on earlier versions before TS6 changed the diagnostic. Closed issue [#181](https://github.com/workos/authkit-nextjs/issues/181) (1.x era) was a related-but-different surfacing of the public-types-export problem.

Leestex:
Closes #422.

## Why

`@workos-inc/node`'s types (`User`, `OauthTokens`, `AuthenticationResponse`) are surfaced into authkit-nextjs's public-API interfaces in [`src/interfaces.ts`](https://github.com/workos/authkit-nextjs/blob/main/src/interfaces.ts) (`Session`, `UserInfo`, `HandleAuthSuccessData`), but `@workos-inc/node` is declared in regular `dependencies`. Any time a consumer's resolved version of `@workos-inc/node` diverges from the version satisfying the transitive `^9.0.0` range, package managers install a duplicate nested copy under `node_modules/@workos-inc/authkit-nextjs/node_modules/@workos-inc/node/`. Consumer code that infers return types from `withAuth()` / `refreshSession()` then hits `TS2883` ("inferred type cannot be named without a reference to 'User' from … This is likely not portable"). Full repro and analysis in #422.

## What

Move `@workos-inc/node` from `dependencies` → `peerDependencies` (and add it to `devDependencies` so the package's own build / typecheck / tests continue to find it). Same pattern as the existing `next`, `react`, `react-dom` peer-dep declarations — the peer-dep contract forces a single resolved copy in the consumer tree by construction.

```diff
   "dependencies": {
     "@sindresorhus/fnv1a": "^3.1.0",
-    "@workos-inc/node": "^9.0.0",
     "iron-session": "^8.0.4",
     "jose": "^5.10.0",
     "path-to-regexp": "^6.3.0",
     "valibot": "^1.2.0"
   },
   "peerDependencies": {
+    "@workos-inc/node": "^9.0.0",
     "next": "^13.5.9 || ^14.2.26 || ^15.2.3 || ^16",
     "react": "^18.0 || ^19.0.0",
     "react-dom": "^18.0 || ^19.0.0"
   },
   "devDependencies": {
     ...
+    "@workos-inc/node": "^9.0.0",
     ...
   }
```

## Verification

Locally on this branch:

- `pnpm install` — clean install, lockfile regenerates as expected (only the dep-tree `@workos-inc/node` block moves; no version churn).
- `pnpm run build` — green (`tsc --project tsconfig.app.json`).
- `pnpm run typecheck` — green (`tsc --project tsconfig.app.json --noEmit && tsc --project tsconfig.test.json --noEmit`).
- `pnpm run lint` — green (`oxlint`, 0 warnings / 0 errors).
- `pnpm run test` — green (21 test files, 349 tests, 0 failures).

End-to-end: cloning a fresh consumer, removing the `resolutions` / dedupe workaround, and pointing at this branch via `pnpm pack` → consumer install resolves a single copy of `@workos-inc/node` regardless of the consumer's pinned version, and the `TS2883` from #422 is gone.

## Compatibility / migration

This is a breaking change for any consumer that doesn't already have `@workos-inc/node` listed in their own `dependencies`. In practice the realistic impact is small — every authkit-nextjs consumer already calls `@workos-inc/node` directly for the parts of the WorkOS SDK that authkit-nextjs doesn't wrap (organization management, user management, FGA, audit logs, vault, etc.). The `chore(deps)!:` prefix on the commit signals the major-bump intent for release-please.

Suggested follow-up release notes (drop into the next major):

> `@workos-inc/node` is now a peer dependency. Consumers must install it directly: `pnpm add @workos-inc/node`. This prevents `TS2883` errors on inferred types from `withAuth()` and related helpers when the consumer pins a specific version of `@workos-inc/node`.

If a softer landing is preferred, a backportable variant is to keep `@workos-inc/node` in both `dependencies` and `peerDependencies` — but Yarn / npm / pnpm don't all dedupe that consistently, and the simpler peer-only declaration matches what's already in place for `next` and `react`. Happy to adjust if maintainers prefer a different shape.


gjtorikian (approved):

gjtorikian:
thanks!

PR #428, opened 2026-05-15
feat: Add feature flags runtime client helper

stanleyphu:
## Summary
- add a memoized helper for the WorkOS Feature Flags runtime client
- document both feature flag evaluation paths: access token claim and runtime client
- add a focused test for runtime client memoization

## Tests
- corepack pnpm@10.28.0 test src/feature-flags.spec.ts
- corepack pnpm@10.28.0 run typecheck
- corepack pnpm@10.28.0 run lint
- corepack pnpm@10.28.0 exec oxfmt --check README.md src/index.ts src/feature-flags.ts src/feature-flags.spec.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,7 +481,11 @@ const featureFlags = getFeatureFlagsRuntimeClient();
 export default async function DashboardPage() {
   const { user, organizationId } = await withAuth({ ensureSignedIn: true });
 
-  await featureFlags.waitUntilReady({ timeoutMs: 5000 });
+  try {
+    await featureFlags.waitUntilReady({ timeoutMs: 5000 });
+  } catch (error) {
+    console.error('Feature flags client failed to initialize:', error);
+  }
 
   const enabled = featureFlags.isEnabled('advanced-analytics', {
     userId: user.id,
--- a/README.md
+++ b/README.md
@@ -452,13 +452,47 @@ export default function MyComponent() {
 }
 ```
 
-### Get the enabled flags for the logged in user
+### Evaluate feature flags
 
-For situations where you need access to the authenticated user's currently active feature flags, use `withAuth` to retrieve the flags from the WorkOS session.
+There are two ways to evaluate feature flags with AuthKit Next.js:
+
+- Use the `feature_flags` access token claim when you want a simple session-scoped list of enabled flags for the signed-in user.
+- Use the Feature Flags runtime client when you want server-side evaluation without storing every active flag in the user's session cookie.
+
+#### Option 1: Use the `feature_flags` claim
+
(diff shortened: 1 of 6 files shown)


src/feature-flags.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { getWorkOS } from './workos.js';
+
+describe('feature flags', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
```
gjtorikian, line 5: I think you'll also need `vi.resetModules()` here. `featureFlagsRuntimeClient` is at the module level, and it's only ever assigned, never cleared, so subsequent tests will keep "growing" the options provided to the client. (Alternatively, you can expose some kind of `close` method in `src/feature-flags.ts` to reset the client, but I think that's a little heavy handed.)

src/feature-flags.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Returns a shared WorkOS Feature Flags runtime client.
+ *
+ * The runtime client keeps feature flag state in sync in the background, so it
+ * should be created once per server process instead of once per request.
+ * Options are only used when the client is created for the first time.
+ */
+export function getFeatureFlagsRuntimeClient(options?: RuntimeClientOptions): FeatureFlagsRuntimeClient {
+  featureFlagsRuntimeClient ??= getWorkOS().featureFlags.createRuntimeClient(options);
```
gjtorikian: src/utils.ts provides a [`lazy<T>()`](https://github.com/workos/authkit-nextjs/blob/5df55e19004fe4a2c4c8023ceb339b806a810efd/src/utils.ts#L41) helper, so we can also be lazy here and reuse it.

README.md
```diff
@@ -452,14 +452,48 @@ export default function MyComponent() {
...
+
+```tsx
+import { getFeatureFlagsRuntimeClient, withAuth } from '@workos-inc/authkit-nextjs';
+
+const featureFlags = getFeatureFlagsRuntimeClient();
+
+export default async function DashboardPage() {
+  const { user, organizationId } = await withAuth({ ensureSignedIn: true });
+
+  await featureFlags.waitUntilReady({ timeoutMs: 5000 });
```
gjtorikian: OOC what happens if we hit the timeout? Do we need to add a `try` / `catch` or something here to handle that, or is it unlikely in practice?

stanleyphu: We throw an error `Error("waitUntilReady timed out")` so we should definitely update the example here with a try-catch. We already do it in our docs: https://workos.com/docs/feature-flags/sdk-integration#3-basic-setup

gjtorikian (approved):

PR #432, opened 2026-05-26
fix: defer PKCE cookie write until AuthKit redirect

nicknisi:
## Summary

- Middleware previously wrote a new `wos-auth-verifier-{hash}` cookie on **every request** without a session. Because each flow produces a uniquely named cookie, repeated page loads (e.g., refreshing while signed out) accumulated cookies until request headers exceeded server limits (HTTP 431).
- Instead of writing the cookie speculatively, the sealed PKCE state is now stashed in internal middleware headers (`x-workos-pkce-state`, `x-workos-authorization-url`). The actual `Set-Cookie` is only emitted when `handleAuthkitHeaders` redirects to the matching AuthKit authorization URL. No redirect → no cookie → no accumulation.
- For backward compatibility with `authkit()` users who forward headers manually without `handleAuthkitHeaders`, the speculative `Set-Cookie` is still included in the raw headers. `handleAuthkitHeaders` strips it on non-redirect responses via `stripPKCESetCookieHeaders`, so the recommended path never accumulates.
- A threshold-based cleanup (`MAX_PKCE_COOKIES=5`) is kept as a safety net for cookies created via server actions (`signIn`, `switchToOrganization`) which use `setPKCECookie` directly.

### DIFF
--- a/src/pkce.ts
+++ b/src/pkce.ts
@@ -102,7 +102,7 @@ export function stripPKCESetCookieHeaders(headers: Headers): void {
   }
 }
 
-function isInitialDocumentRequest(request: NextRequest): boolean {
+export function isInitialDocumentRequest(request: NextRequest): boolean {
   const accept = request.headers.get('accept') || '';
   const isDocumentRequest = accept.includes('text/html');
   const isRSCRequest = request.headers.has('RSC') || request.headers.has('Next-Router-State-Tree');
--- a/src/pkce.ts
+++ b/src/pkce.ts
@@ -1,12 +1,17 @@
 import fnv1a from '@sindresorhus/fnv1a';
 import { unsealData } from 'iron-session';
 import { cookies } from 'next/headers';
+import { NextRequest } from 'next/server';
 import * as v from 'valibot';
 import { getPKCECookieOptions } from './cookie.js';
 import { WORKOS_COOKIE_PASSWORD } from './env-variables.js';
 import { State, StateSchema } from './interfaces.js';
 
 export const PKCE_COOKIE_NAME = 'wos-auth-verifier';
+export const PKCE_STATE_HEADER = 'x-workos-pkce-state';
+export const PKCE_AUTHORIZATION_URL_HEADER = 'x-workos-authorization-url';
+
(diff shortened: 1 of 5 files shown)


nicknisi:
@greptileai

src/pkce.ts
```diff
@@ -43,6 +48,81 @@ export async function setPKCECookie(sealedState: string): Promise<void> {
...
+  headers.delete('Set-Cookie');
+
+  for (const setCookieHeader of setCookieHeaders) {
+    if (!isPKCESetCookieHeader(setCookieHeader)) {
+      headers.append('Set-Cookie', setCookieHeader);
+    }
+  }
+}
+
+function isInitialDocumentRequest(request: NextRequest): boolean {
```
gjtorikian: This seems to be the same as [this pre-existing function](https://github.com/workos/authkit-nextjs/blob/591999b5c22d29a11b3211adb7e150ed8cb37539/src/session.ts#L82) -- maybe we can extract them out into one source?

gjtorikian (approved):
minor comment that i think should be addressed, but because the suggestion is quick to implement, and to reduce the review back-and-forth, approving

src/pkce.ts
```diff
@@ -1,12 +1,17 @@
...
 import { unsealData } from 'iron-session';
 import { cookies } from 'next/headers';
+import { NextRequest } from 'next/server';
 import * as v from 'valibot';
 import { getPKCECookieOptions } from './cookie.js';
 import { WORKOS_COOKIE_PASSWORD } from './env-variables.js';
 import { State, StateSchema } from './interfaces.js';
 
 export const PKCE_COOKIE_NAME = 'wos-auth-verifier';
+export const PKCE_STATE_HEADER = 'x-workos-pkce-state';
```
cmatheson, line 11: i think i see why/how we ended up with this problem, but imo we should:

1. deprecate `authorizationUrl` from `authkit`
2. add `signIn`/`signUp` helpers like authkit-js has (or `redirectToAuthorizationUrl`/etc)

I think this way we always know when to set the cookie and we have less state to track, our customers don't need to worry about setting the cookie on their side/etc.

PR #438, opened 2026-06-10
test: pin callback redirect to the app origin

nicknisi:
## What

Adds regression tests to `authkit-callback-route.spec.ts` asserting that the post-auth redirect always stays on the application's own origin, regardless of what `returnPathname` contains.

Hostile inputs covered:
- `javascript:alert(document.domain)`
- `https://evil.com/phishing`
- `//evil.com/phishing` (protocol-relative)
- `/\evil.com/phishing` (backslash → slash normalization)
- `https:/evil.com`

Each asserts the resulting `Location` has `protocol === 'http:'`, `origin === 'http://example.com'`, and `hostname === 'example.com'`.

## Why

A customer report flagged that the OAuth `state` pattern (encoding a return URL) is an open-redirect / `javascript:`-URI footgun when consumers redirect to it unvalidated.

`@workos-inc/authkit-nextjs` is **already safe by construction**: `returnPathname` is read only from the iron-session-sealed PKCE cookie (never raw URL `state`), and the callback copies only `pathname` + `search` onto the app's own origin (`authkit-callback-route.ts`). This neutralizes `javascript:`, absolute, and protocol-relative URLs.

These tests pin that invariant so a future refactor that started honoring the full URL would fail loudly. No production code changes.

## Notes

A `https:/evil.com` input lands at `http://example.com/evil.com` — a harmless path segment on our own origin, not a redirect to evil.com. The tests assert on origin/host rather than string-matching, to reflect the real security property.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/authkit-callback-route.spec.ts
+++ b/src/authkit-callback-route.spec.ts
@@ -243,6 +243,7 @@ describe('authkit-callback-route', () => {
       const appOrigin = 'http://example.com';
       const hostileReturnPathnames = [
         'javascript:alert(document.domain)',
+        'data:text/html,<script>alert(1)</script>',
         'https://evil.com/phishing',
         '//evil.com/phishing',
         '/\\evil.com/phishing',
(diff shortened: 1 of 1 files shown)


willporter-workos (approved):
These tests make sense to me

gjtorikian (approved):

PR #446, opened 2026-06-28
feat: add recent-auth checking (checkRecentAuth + useRecentAuth) and maxAge step-up

imkesin:
Surface the access token's `auth_time` claim so apps can enforce step-up / reauthentication.

## What's in here

- **`checkRecentAuth` server function** — data-only recency check that fails closed (`isStale: true` when `auth_time` is missing or there's no user). It never redirects, so it's safe to call as the enforcement step inside a sensitive server action or in a server component where you decide what to do.
- **`useRecentAuth` client hook** — presentation-only recency state from claims already in client memory.
- **`evaluateRecentAuth`** — shared internal helper (in `utils.ts`) backing both. Fails closed on a missing/non-finite `auth_time`, treats a future `auth_time` (clock skew) as recent, and uses an inclusive `maxAge` boundary.
- Added `auth_time` to `JWTPayload`.

## `maxAge` step-up (OIDC `max_age`)

- `getSignInUrl` / `getSignUpUrl` / `getAuthorizationUrl` now accept `maxAge`, forwarding OIDC `max_age` so the IdP forces a reauth when the most recent auth is older. This is the redirect half of the step-up loop that `checkRecentAuth` gates.
- The `GetAuthURLOptions.maxAge` field is **version-gated via a conditional type** derived from the installed SDK (`Parameters<WorkOS['userManagement']['getAuthorizationUrl']>[0]`). On `@workos-inc/node` < 10.6 it resolves to `never` — unsettable at compile time rather than advertised and silently dropped at runtime. No false promise, no silent no-op.

## Dependencies

- **Non-breaking.** The peer range is unchanged (`@workos-inc/node` `^9.0.0 || ^10.0.0`); the conditional type self-gates per consumer.
- Dev dependency bumped to `@workos-inc/node` `^10.7.0` so the build type-checks the `maxAge` passthrough.

## Notes for reviewers

- This package is standalone (it does not depend on `@workos/authkit-session`), so `checkRecentAuth` reads `auth_time` via the existing `withAuth` + `getTokenClaims` path. That read is decode-only, consistent with `withAuth`/`getTokenClaims` — the iron-session seal (verified/refreshed in middleware) is the integrity boundary throughout the library.
- Parallels workos/authkit-tanstack-start#112.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,6 +31,33 @@ export function errorResponseWithFallback(errorBody: { error: { message: string;
       });
 }
 
+type EvaluateRecentAuthParameters = {
+  authTime: unknown;
+  maxAgeSeconds: number;
+  nowSeconds: number;
+};
+
+/**
+ * Evaluate whether an authentication is recent enough.
+ *
+ * Fails closed: a missing or non-finite `authTime` is reported as stale. A
+ * future `authTime` (clock skew) is treated as recent rather than stale, and
+ * the `maxAge` boundary is inclusive.
+ */
+export function evaluateRecentAuth({ authTime, maxAgeSeconds, nowSeconds }: EvaluateRecentAuthParameters) {
+  if (typeof authTime !== 'number' || !Number.isFinite(authTime)) {
+    return {
+      authenticatedAt: null,
+      isStale: true,
+    } as const;
+  }
+
+  return {
+    authenticatedAt: new Date(authTime * 1000),
+    isStale: nowSeconds - authTime > maxAgeSeconds,
+  } as const;
+}
+
 /**
  * Returns a function that can only be called once.
  * Subsequent calls will return the result of the first call.
--- a/README.md
+++ b/README.md
@@ -512,6 +512,51 @@ const { user, loading } = useAuth({ ensureSignedIn: true });
 
 Enabling `ensureSignedIn` will redirect users to AuthKit if they attempt to access the page without being authenticated.
 
+### Re-authentication
+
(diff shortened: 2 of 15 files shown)


src/utils.ts
```diff
@@ -31,6 +31,33 @@ export function errorResponseWithFallback(errorBody: { error: { message: string;
...
+
+/**
+ * Evaluate whether an authentication is recent enough.
+ *
+ * Fails closed: a missing or non-finite `authTime` is reported as stale. A
+ * future `authTime` (clock skew) is treated as recent rather than stale, and
+ * the `maxAge` boundary is inclusive.
+ */
+export function evaluateRecentAuth({ authTime, maxAgeSeconds, nowSeconds }: EvaluateRecentAuthParameters) {
+  if (typeof authTime !== 'number' || !Number.isFinite(authTime)) {
```
gjtorikian, line 48: Since `NaN` is a number, I think `maxAgeSeconds` needs a similar check here:

```
if (typeof authTime !== 'number' || !Number.isFinite(authTime) || !Number.isFinite(maxAgeSeconds)) {
  ...
}
```

Because otherwise, `isStale` will always be false (`nowSeconds - authTime > NaN == false`).

Bonus points: add `it('fails closed when maxAge is not finite')` to utils.spec.ts

gjtorikian (commented):
The README here has a pretty thorough Usage section -- we may want to also add some kind of "Enforcing recent authentication (step-up)"  guide here.

src/utils.ts
imkesin: Does it change your perspective if `maxAgeSeconds` is always the user provided value? I also don't check `nowSeconds` on the way into this function.

gjtorikian (approved):

PR #453, opened 2026-07-16
fix: proactively refresh near-expiry access tokens in the proxy middleware

### LINKED ISSUE #452: Middleware has no pre-expiry refresh buffer, so a near-expiry token reaches server-side fetches expired (intermittent 500s after idle)
**Describe the bug**

There is a refresh-timing asymmetry between the client token store and the middleware:

- The **client** (`src/components/tokenStore.ts`) refreshes *before* expiry: `TOKEN_EXPIRY_BUFFER_SECONDS = 60` (30s for tokens with a <=300s lifetime). `parseToken` marks a token `isExpiring` when `exp < now + buffer`.
- The **middleware** (`src/session.ts` `updateSession` -> `verifyAccessToken`) refreshes *only reactively*. `verifyAccessToken` is a bare `jwtVerify` with no `clockTolerance` and no pre-expiry buffer, so it treats any token where `exp > now` as valid and passes it straight through:

```js
const hasValidSession = await verifyAccessToken(session.accessToken);
if (hasValidSession) {
  newRequestHeaders.set(sessionHeaderName, request.cookies.get(cookieName).value);
  // ...token forwarded to withAuth() as-is, no refresh
}
```

The refresh path (`authenticateWithRefreshToken`) runs only once `jwtVerify` *fails*, i.e. the token is already past `exp`.

Consequence: a token with a few seconds of life left clears the middleware, is handed to `withAuth()`, and is passed to a **server-side** call (a Server Component `fetchQuery`, an API call to another service, etc.). By the time that downstream service validates the JWT, render latency + the network round-trip + clock skew have pushed it past `exp`, and it is rejected. The client never hits this because of its 60s head start.

In production this surfaces as a dead-end 500: the downstream auth error is thrown during the RSC render, and Next.js redacts render errors, so an `error.tsx` boundary cannot classify it as an auth failure and recover.

This is distinct from #289 (client `useAccessToken` staleness after tab inactivity, fixed in v2.5.0 with focus/visibility listeners) which is the client side only; the middleware still has no buffer.

**To Reproduce**

1. Protect a route with `authkitMiddleware({ middlewareAuth: { enabled: true } })`.
2. In a Server Component on that route, call a downstream service that validates the WorkOS access token server-side (e.g. `fetchQuery(..., { token: accessToken })`), where the service rejects an expired token.
3. Let the access token drift to within a few seconds of `exp` (easily reached after tab inactivity / sleep throttling, when the client's background refresh timer is throttled). Navigate.
4. Intermittently, the middleware forwards the near-expiry token, it expires in transit, the downstream call rejects, and the render 500s.

**Expected behavior**

The middleware should refresh a near-expiry (but still valid) token proactively, mirroring the client buffer, so a token handed to `withAuth()` / server-side fetches always has comfortable remaining lifetime.

Proposed fix, either:
- Mirror the client buffer in the middleware: refresh when `exp < now + buffer` (same 60s / 30s-for-short-lived thresholds as `tokenStore`) instead of only after `jwtVerify` fails.
- Or expose a config option (e.g. `refreshBufferSeconds` on `authkitMiddleware`) so consumers can opt into a pre-expiry refresh window.

Today there is no supported way to add this. As a stopgap we wrap the middleware and re-seal a freshly refreshed session into the request cookie before AuthKit runs, but that replicates internals (the iron-session seal shape, cookie name, `getCookieOptions`) and is fragile across versions. A first-class buffer or hook would let us drop it.

**Screenshots**

n/a: the failure is a server-side render 500; there is nothing visual beyond the framework error page.

- OS: Linux (server-side; Vercel, middleware/`proxy.ts` on the Node runtime)
- Browser: n/a (server-side render)
- authkit-nextjs version: 4.2.0
- Next.js version: 16.x

**Additional context**

Downstream service in our case is Convex (`fetchQuery(..., { token })`), but the bug is provider-agnostic: it affects any server-side consumer of `accessToken` that validates `exp`.

wouteth:
Fixes #452.

## Problem

There is a refresh-timing asymmetry between the client token store and the proxy/middleware:

- The client (`src/components/tokenStore.ts`) refreshes access tokens **before** expiry: 60 seconds early, or 30 seconds for tokens with a total lifetime of 5 minutes or less.
- The middleware (`updateSession` -> `verifyAccessToken`) refreshes **only reactively**, once the token is already past `exp`.

A token with a few seconds of life left passes the middleware unchanged, is handed to `withAuth()` and then to a server-side consumer (a Server Component fetch, an API call to a service that validates `exp`), and expires in transit due to render latency, network round trips, or clock skew. In production this surfaces as an intermittent 500 during the RSC render (Next.js redacts render errors, so an `error.tsx` boundary cannot classify it as an auth failure and recover). Full details and a repro in #452.

## Fix

`updateSession` now treats a still-valid token that is within a refresh buffer of its expiry as needing refresh, so the existing (already correct) refresh path runs for it: new token pair, re-sealed cookie, `x-workos-session` header, `Set-Cookie`.

- The default buffer mirrors the client token store exactly: 60 seconds, or 30 seconds for tokens with a total lifetime of <= 5 minutes, so the SDK is self-consistent out of the box.
- A new `refreshBufferSeconds` option on `authkitProxy` / `authkit()` tunes the buffer; `0` disables proactive refresh and restores today's reactive-only behaviour. (If you would prefer the feature to be opt-in instead of opt-out, flipping the default to `0` is a one-line change; happy to do that.)

**Failed proactive refresh is not fatal.** Refresh tokens are single-use, so under parallel navigation requests inside the buffer window, losers of the refresh race get a WorkOS error. The existing catch branch would delete the session cookie and redirect to sign-in, which would turn a benign race into a logout. Instead, when the current token is still valid, the request is served with it (the race winner has already persisted the rotated session). `onSessionRefreshError` is intentionally not fired in that fallback, since no user-visible auth failure occurred.

Cost note: this adds one `authenticateWithRefreshToken` call per token lifetime (in the buffer window), not per request.

## Testing

- 8 new tests: buffer boundary in/out, the 30-second short-lived-token tier, custom `refreshBufferSeconds`, `0` opt-out, the failed-refresh fallback (token served, cookie NOT deleted, no `onSessionRefreshError`), `onSessionRefreshSuccess` on proactive refresh, and option threading through `updateSessionMiddleware` / `authkitProxy`.
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run format:check`, `pnpm run build`, `pnpm test` all green (393 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -331,16 +331,21 @@ async function updateSession(
     };
   } catch (e) {
     if (isExpiring) {
-      // The current token is still valid, so a failed proactive refresh is not fatal.
-      // Refresh tokens are single-use, so a concurrent request in the same buffer
-      // window may have already rotated this one; that request has persisted the new
-      // session cookie. Serve this request with the current token instead of
-      // destroying the session.
-      if (options.debug) {
-        console.log('Proactive refresh failed. Serving request with the still-valid access token.', e);
-      }
+      // A failed proactive refresh is not fatal while the current token is still
+      // valid. Refresh tokens are single-use, so a concurrent request in the same
+      // buffer window may have already rotated this one; that request has persisted
+      // the new session cookie. Serve this request with the current token instead of
+      // destroying the session. Re-check validity here: the token may have expired
+      // during the refresh round trip, in which case fall through to the
+      // delete-cookie path below.
+      const { exp } = decodeJwt(session.accessToken);
+      if (typeof exp === 'number' && exp > Math.floor(Date.now() / 1000)) {
+        if (options.debug) {
+          console.log('Proactive refresh failed. Serving request with the still-valid access token.', e);
+        }
 
-      return respondWithCurrentToken();
+        return respondWithCurrentToken();
+      }
     }
 
     if (options.debug) {
--- a/src/session.ts
+++ b/src/session.ts
@@ -95,6 +95,7 @@ async function updateSessionMiddleware(
   redirectUri: string,
   signUpPaths: string[],
   eagerAuth = false,
+  refreshBufferSeconds?: number,
 ) {
   if (!redirectUri && !WORKOS_REDIRECT_URI) {
     throw new Error('You must provide a redirect URI in the AuthKit middleware or in the environment variables.');
@@ -140,6 +141,7 @@ async function updateSessionMiddleware(
     redirectUri,
     screenHint: getScreenHint(signUpPaths, request.nextUrl.pathname),
     eagerAuth,
+    refreshBufferSeconds,
   });
 
   // Record the sign up paths so we can use them later
@@ -210,12 +212,13 @@ async function updateSession(
   }
 
   const hasValidSession = await verifyAccessToken(session.accessToken);
+  const isExpiring = hasValidSession && isTokenExpiring(session.accessToken, options.refreshBufferSeconds);
 
   const cookieName = WORKOS_COOKIE_NAME || 'wos-session';
 
   applyCacheSecurityHeaders(newRequestHeaders, request, session);
 
-  if (hasValidSession) {
+  const respondWithCurrentToken = (): AuthkitResponse => {
     newRequestHeaders.set(sessionHeaderName, request.cookies.get(cookieName)!.value);
 
     const {
@@ -253,13 +256,19 @@ async function updateSession(
       },
       headers: newRequestHeaders,
     };
+  };
+
+  if (hasValidSession && !isExpiring) {
+    return respondWithCurrentToken();
   }
 
   try {
     if (options.debug) {
       // istanbul ignore next
       console.log(
-        `Session invalid. ${session.accessToken ? `Refreshing access token that ends in ${session.accessToken.slice(-10)}` : 'Access token missing.'}`,
+        isExpiring
+          ? `Session expiring soon. Proactively refreshing access token that ends in ${session.accessToken.slice(-10)}`
+          : `Session invalid. ${session.accessToken ? `Refreshing access token that ends in ${session.accessToken.slice(-10)}` : 'Access token missing.'}`,
       );
     }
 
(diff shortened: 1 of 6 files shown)


src/session.ts
```diff
@@ -321,6 +330,19 @@ async function updateSession(
       headers: newRequestHeaders,
     };
   } catch (e) {
+    if (isExpiring) {
```
gjtorikian, line 333: I think we should re-check the remaining validity in the catch before falling back:

```typescript
if (isExpiring) {
  const { exp } = decodeJwt(session.accessToken);
  if (typeof exp === 'number' && exp > Math.floor(Date.now() / 1000)) {
    return respondWithCurrentToken();
  }
  // fall through to the existing delete-cookie path
}
```

gjtorikian (changes requested):
Overall looks quite good. In addition to the comment above, consider adding a note in the README's proactive-refresh section that parallel requests inside the buffer window may each pay a failed-refresh round trip before being served.

We can keep this behavior default-on.

src/session.ts
wouteth: Good catch, done in 8521962. The catch now re-checks `exp` against the current time and only serves the current token if it survived the refresh round trip; otherwise it falls through to the existing delete-cookie path (and `onSessionRefreshError` fires as before). Added a test that simulates the token expiring mid-refresh.

wouteth:
Thanks for the review! Both points addressed in 8521962:

- The failed-refresh fallback now re-checks remaining validity before serving the current token; if it expired during the refresh attempt it falls through to the delete-cookie path. New test covers the mid-refresh expiry case.
- Added a note to the README's proactive-refresh section that parallel requests inside the buffer window may each pay a failed-refresh round trip before being served.

Kept the behavior default-on as discussed. Full suite green (394 tests), typecheck/lint/format all pass.

gjtorikian:
thanks!

